PR #1, opened 2017-11-15
ddi: do not assume arguments are correct

Aearsis:
When an uspace app requests to map 0 bytes, dmamem_map_anonymous calls
frame_alloc, which asserts the size is not zero, causing kernel panic.


jermar:
Thanks.

PR #2, opened 2017-12-06
Fix toolchain build broken on macOS 10.13

nebulabox:
A minor gdb patch, use standard 'fmin' instead of 'min'.
Change the toolchain.sh,  get SRCDIR by 'pwd -P' instead of 'readlink'. The 'readlink' on macOS is not same as the GNU's 'readlink'.


le-jzr:
Hi @nebulabox, are you sure the patch is necessary? It would be nice if you could check whether a more recent version of the upstream package works. It would be better to move to a newer version than to patch old bugs.

nebulabox:
You should check the build on lastest macOS instead of just thinking it.

PR #3, opened 2017-12-06
Update .gitignore to ignore toolchain build temp files.

nebulabox:
ignore the download files and build temp folder.


vhotspur:
Thanks!

PR #4, opened 2017-12-06
Run make with -j to improve the build speed.

nebulabox:
Seems binutils, gcc and gdb support run make with -j, so build the toolchain simultaneously.


vhotspur:
Nice patch, thanks. But would it be possible to make this somehow optional? You can build for multiple architectures in parallel and this could have adverse effects. I would also check that `CPUCOUNT` is never empty and set it to `1` then (e.g. when the Python mini-script fails for whatever reason).

le-jzr:
I would _strongly_ prefer we use `nproc` instead of a python script. Is that not working for you, @nebulabox?

nebulabox:
`nproc` is not available for macOS but python is default available for macOS and most of Linux.

PR #5, opened 2017-12-08
Return phone handle in SYS_IPC_CONNECT_KBOX separately from error code.

le-jzr:
@jermar I'm not sure how to properly deal with resources on failure. Could you have a look?

### DIFF
--- a/kernel/generic/src/ipc/kbox.c
+++ b/kernel/generic/src/ipc/kbox.c
@@ -231,11 +231,26 @@ int ipc_connect_kbox(task_id_t taskid, cap_handle_t *out_phone)
 		return ENOENT;
 	}
 	
-	if (task->kb.finished != false) {
+	if (task->kb.finished) {
 		mutex_unlock(&task->kb.cleanup_lock);
 		return EINVAL;
 	}
 	
+	/* Create a kbox thread if necessary. */
+	if (task->kb.thread == NULL) {
+		thread_t *kb_thread = thread_create(kbox_thread_proc, NULL, task,
+		    THREAD_FLAG_NONE, "kbox");
+		
+		if (!kb_thread) {
+			mutex_unlock(&task->kb.cleanup_lock);
+			return ENOMEM;
+		}
+		
+		task->kb.thread = kb_thread;
+		thread_ready(kb_thread);
+	}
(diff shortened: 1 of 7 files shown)


kernel/generic/src/ipc/kbox.c
```diff
@@ -249,13 +250,15 @@ int ipc_connect_kbox(task_id_t taskid)
...
-		return phone_handle;
+		*out_phone = phone_handle;
+		return EOK;
 	}
 	
 	/* Create a kbox thread */
 	thread_t *kb_thread = thread_create(kbox_thread_proc, NULL, task,
 	    THREAD_FLAG_NONE, "kbox");
 	if (!kb_thread) {
+		// FIXME: Shouldn't we clean up phone_handle?
```
jermar: With the present code the phone would be cleanup up during ipc_cleanup() when the task is destroyed. If you want to clean-up sooner, which may be a good thing, you could either mimic steps from phone_cap_cleanup_cb() (but there might be some calls sent over it already) or you could swap the ordering of kbox thread creation and phone allocation so that the phone gets allocated when the function cannot fail for other reasons.

kernel/generic/src/ipc/sysipc.c
```diff
@@ -900,18 +900,24 @@ sysarg_t sys_ipc_irq_unsubscribe(sysarg_t cap)
...
+	cap_handle_t phone;
 	int rc = copy_from_uspace(&taskid, uspace_taskid, sizeof(sysarg64_t));
-	if (rc != 0)
-		return (sysarg_t) rc;
+	if (rc == EOK) {
+		rc = ipc_connect_kbox((task_id_t) taskid, &phone);
+	}
+	if (rc == EOK) {
+		rc = copy_to_uspace(uspace_phone, &phone, sizeof(cap_handle_t));
+		// FIXME: Clean up phone on failure.
```
jermar: sys_ipc_hangup() comes to mind here. Even though I'm planning changes in this area that would destroy the phone capability non-lazily.

le-jzr:
@jermar Thanks for the feedback! I reordered the thread creation and piggybacked on sys_ipc_hangup() for cleanup.

PR #8, opened 2017-12-20
"Obviously harmless" error handling tweaks.

le-jzr:
This is a subset of changes necessary to introduce `errno_t` as a typechecking aid. A few of these changes are bug fixes, but most are simple modifications with no semantic effect (e.g. converting EOK to 0 and vice versa, splitting `int rc` into two separate variables, separating `rc` from variables declared on the same line, etc).

The intention is that all the changes in this set are obviously harmless, i.e. one can trivially see they won't cause any regressions, without having to look up surrounding code. Their actual _correctness_ may not be obvious at a glance, but I'm fairly confident they are.

I'll leave this open until I get enough sleep to give this another thorough review (probably late afternoon tomorrow), then I'll merge this as is unless I discover new mistakes.

### DIFF
--- a/uspace/app/bdsh/cmds/builtins/batch/batch.c
+++ b/uspace/app/bdsh/cmds/builtins/batch/batch.c
@@ -84,7 +84,7 @@ int cmd_batch(char **argv, cliuser_t *usr)
 			continue_despite_errors = true;
 	}
 
-	int rc = 0;
+	int rc = EOK;
 	FILE *batch = fopen(argv[1], "r");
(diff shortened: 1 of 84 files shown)


jxsvoboda:
LGTM. Note that we will also need to fix all the places in code docs where it says we return something like 'EOK or negative error code' to say just 'EOK or error code'

le-jzr:
Reviewed it once more and didn't spot any problem except for one older bug. Going ahead with the merge. :)

PR #9, opened 2017-12-21
The "not-so-obvious" error handling tweaks.

le-jzr:
Similar as the previous "obvious" tweaks, except fewer in number and not as obvious this time.
Reviewing some of these changes may require spare brainpower.

Same deal as the previous one: I'll leave this open for comments until late tomorrow, then I'll do one final sweep and merge the changes.

### DIFF
--- a/uspace/app/init/init.c
+++ b/uspace/app/init/init.c
@@ -207,7 +207,7 @@ static int srv_startl(const char *path, ...)
 		printf("%s: Server %s failed to start (exit code %d)\n", NAME,
 		    path, retval);
 	
-	return retval;
+	return retval == 0 ? EOK : EPARTY;
 }
 
 static int console(const char *isvc, const char *osvc)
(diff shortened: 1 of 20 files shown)


uspace/app/init/init.c
```diff
@@ -207,7 +207,7 @@ static int srv_startl(const char *path, ...)
 		printf("%s: Server %s failed to start (exit code %d)\n", NAME,
 		    path, retval);
 	
-	return retval;
+	return retval == 0 ? EOK : EPARTY;
```
jermar: EPARTY was meant as a special error code returned only from the kernel so that it is possible to distinguish one particular error condition from whatever the recipient may return to the sender. I know it is not enforced now other than by convention, but I think that we should not proliferate EPARTY to other places.

le-jzr: Ah, good to know. Which error code do you think would fit best for the case of "child task exited unsuccessfully"?

jermar: I solved this by introducing the IPC_CALL_AUTO_REPLY flag which can only be set by the kernel.

PR #11, opened 2018-01-06
Integrate ports into the main tree as a submodule.

le-jzr:
A stripped down version of hsct.sh is added to tools, and Makefiles
gain new targets that implement part of hsct's prior functionality.

This arrangement improves usability and maintainability of ports,
reduces the likelihood of ports being broken by mainline changes,
and partially solves the issue with build logic being duplicated
in multiple places.

It is newly possible to build individual ports using `make ports-NAME`,
and all ports using `make ports-all`. This also installs the outputs to
`uspace/dist`, so ideally, simple `make ports-all && make` would create
an image with all ports included. This currently doesn't work only
because some ports fail to build.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "ports"]
+	path = uspace/ports
+	url = git@github.com:HelenOS/harbours.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "ports"]
 	path = uspace/ports
-	url = git@github.com:HelenOS/harbours.git
+	url = https://github.com/HelenOS/harbours.git
--- a/Makefile
+++ b/Makefile
@@ -60,6 +60,9 @@ kernel: common
 uspace: common
 	$(MAKE) -r -C uspace PRECHECK=$(PRECHECK)
 
+ports-%: common
+	$(MAKE) -r -C uspace $@ PRECHECK=$(PRECHECK)
+
 precheck: clean
 	$(MAKE) -r all PRECHECK=y
 
--- a/tools/autotool.py
+++ b/tools/autotool.py
@@ -614,6 +614,7 @@ def main():
 					print_error(["Toolchain for target is not installed, or CROSS_PREFIX is not set correctly."])
 				path = "%s/%s/bin" % (cross_prefix, platform)
 		
+		common['TARGET'] = target
 		prefix = "%s-" % target
 		
 		# Compiler
--- a/tools/hsct.sh
+++ b/tools/hsct.sh
@@ -1,0 +1,399 @@
+#!/bin/sh
+
+#
+# Copyright (c) 2013-2017 Vojtech Horky
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#
+# - Redistributions of source code must retain the above copyright
+#   notice, this list of conditions and the following disclaimer.
+# - Redistributions in binary form must reproduce the above copyright
+#   notice, this list of conditions and the following disclaimer in the
+#   documentation and/or other materials provided with the distribution.
+# - The name of the author may not be used to endorse or promote products
+#   derived from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE AUTHOR ``AS IS'' AND ANY EXPRESS OR
+# IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+# OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+# IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT,
+# INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT
+# NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF
+# THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#
+
+# Calling harbour functions:
+#    These functions are always called in a subshell to "guard" them a little
+# bit (variables set by the harbour, cd into package directory).
+# 
+# Notice on usage of set -o errexit (set -e) 
+#    We want to use that option for the harbour scripts to get rid of the
+# "|| return 1" at the end of each line.
+#    Obvious solution is to wrap the call like this:
+#       (  set -o errexit; build ) || { hsct_error "..."; return 1; }
+#    This doesn't work because the whole subshell is then part of a ||
+# operand and thus the set -e is ignored (even if it is a subshell).
+# See https://groups.google.com/d/msg/gnu.bash.bug/NCK_0GmIv2M/y6RQF1AWUQkJ
+#    Thus, we need to use the following template to get past this:
+#       ( set -o errexit; build; exit $? );
+#       [ $? -eq 0 ] || { hsct_error "..."; return 1; }
+#
+# Also notice that we never ever call exit from the top-most shell when
+# leaving after an error. That is to prevent terminating user shell when
+# this script is sourced ("env" command). It complicates the error handling
+# a bit but it is more reliable than trying to guess whether we are running
+# in a subshell or not.
+#
+
+# Necessary because this script is called from a Makefile.
+# The parent Makefile's options would otherwise pollute the build environment.
+# TODO: It would be helpful to extract the -j option out of this.
+unset MAKEFLAGS
+unset GNUMAKEFLAGS
+
+# Print short help.
+# Does not exit the whole script.
+hsct_usage() {
+	echo "Usage:"
+	echo " $1 action [package]"
+	echo "    Action can be one of following:"
+	echo "       clean     Clean built directory."
+	echo "       fetch     Fetch sources (e.g. download from homepage)."
+	echo "       build     Build given package."
+	echo "       package   Save installable files to allow cleaning."
+	echo "       install   Install to uspace/dist of HelenOS."
+	echo "       archive   Create tarball instead of installing."
+	echo " $1 help"
+	echo "    Display this help and exit."
+}
+
+# Print high-level information message.
+hsct_info() {
+	echo ">>>" "$@" >&2
+}
+
+# Print lower-level information message (additional info after hsct_info).
+hsct_info2() {
+	echo "     ->" "$@" >&2
+}
+
+# Print information message from HARBOUR script.
+msg() {
+	hsct_info "$@"
+}
+
+# Print high-level error message.
+hsct_error() {
+	echo "[hsct]:" "Error:" "$@" >&2
+}
+
+# Print additional details to the error message.
+hsct_error2() {
+	echo "[hsct]:" "  ->  " "$@" >&2
+}
+
+# Run a command but print it first.
+hsct_run_echo() {
+	echo -n "[hsct]: "
+	for ___i in "$@"; do
+		echo -n "$___i" | sed -e 's#"#\\"#g' -e 's#.*#"&" #'
+	done
+	echo
+	"$@"
+}
+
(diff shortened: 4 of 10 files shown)


le-jzr:
I originally planned to utilize the specsfile a bit more than it currently is, so it might currently seem a bit weird in how its done "halfway". I expect that future improvements will simplify things.

vhotspur:
Thanks for the heads-up.

> This arrangement improves usability and maintainability of ports,
> reduces the likelihood of ports being broken by mainline changes,
> and partially solves the issue with build logic being duplicated
> in multiple places.

I would say the build logic was just duplicated :-).

I understand the motivation but I am not sure this is the best way to do it. This helps only in the single case when you are doing substantial changes in `libposix` or `libc`. Otherwise you are more interested in having a pre-built binary as not to waste your time. For that you can download the tarball from [CI](http://ci.helenos.org) or use the `pkg` utility inside running instance.

> It is newly possible to build individual ports using `make ports-NAME`,
> and all ports using `make ports-all`. This also installs the outputs to
> `uspace/dist`, so ideally, simple `make ports-all && make` would create
> an image with all ports included. This currently doesn't work only
> because some ports fail to build.

I think [CI script](https://github.com/HelenOS/ci) does the "build everything" part a bit better especially because of more fine-grained parallelism. But again, if you think this is more convenient, feel free to merge.

To sum it up: I think this pollutes the HelenOS tree more than it deserves but if you believe it helps you in streamlining `libposix` then go ahead :-). For the record I think Git submodules are bad in general and they would eventually bite us. Furthermore here you bind together repositories that were supposed to exist separately (i.e. it should be possible to use upstream Harbours on HelenOS feature branch and vice-versa).

le-jzr:
> This helps only in the single case when you are doing substantial changes in `libposix` or `libc`.
It does. That's the main motivation.

> I think CI script does the "build everything" part a bit better
It does, but running CI on local changes is somewhat laborious. My intention is to allow the "build everything" without any setup at all (analogous in simplicity to `make check`), and I think this does the job well. I break ports often enough to appreciate the simplicity.

> I think this pollutes the HelenOS tree more than it deserves but if you believe it helps you in streamlining libposix then go ahead :-).

I do believe that. :-) Out of curiosity, how would you, in this regard, compare current harbour, tools/xcw, and this change? IMO current harbour pollutes mainline by its very nature as "something that breaks if you change mainline, and it's our problem to keep it fixed". In this sense, the entanglement is already at maximum.

> I think Git submodules are bad in general and they would eventually bite us.
Can you elaborate on this? I consider them a very helpful tool, so I'd like to hear why you think them troublesome.

jxsvoboda:
How about having make ports-xxxx clone the harbours into a subdirectory instead? Like the cstyle checking works right now? I think it makes sense for the cstyle checker, since it's an independent project and we rather want to track the latest version than to fix on a particular version. The same thing could make sense for harbours, too.

We should add unit tests to screen out most problems intruduced with libc/libposix that could break ports.

le-jzr:
I don't think that makes sense. Tracking the latest version of anything is a bad idea, as you have no control over what gets pushed into the third party repository. Malware pushed into such a repository via a compromised developer account would automatically propagate to all your users. This is not an academic concern either, it has happened with javascript libraries in the past.

With ports, security is not a concern (if one repo is compromised, both are), but at this moment, the repositories are nowhere close to being independent. Yes, it's the ideal we should strive for, but it's not currently consistent with how things are.

le-jzr:
Also, (it only occured to me now), pulling the latest revision makes it impossible to bisect for port regressions. Anchoring to a specific remote revision is necessary for history to be meaningful.

jermar:
I like the idea that presently the HelenOS repository does not know anything about coastline and, with the exception of libposix and some other minor offenders, the POSIX world. On the other hand, the use of modules for coastline makes it possible to integrate coastline into HelenOS in a relatively clean way. I say relatively because unfortunately the PR introduces enough coastline scaffolding to make the integration seem more intrusive.

le-jzr:
Yeah, it's a compromise between convenience and purity. The plan is to gradually improve the libraries and reduce the scaffolding, eventually reaching a point where we have a stable API that third-party code can depend on.

le-jzr:
Now that I think about it, we could split the "mainline" into several independent repositories, (e.g. one for kernel, one for uspace, one for the tools, one for ports) and have the main repository serve simply as a convenient integration point. Something to think about.

le-jzr:
Oh, also, it might not be clear from this pull request, but the intention is for the scripts in the harbours repo to be removed. Otherwise this would just add to the maintanance burden.

le-jzr:
Basically, the problem I'm solving is that current `hsct.sh` digs into internal details of the build scripts to figure out which flags to use, and then plays a guessing game about which flags to keep and which to filter out. That's one of the worst possible designs. It's HelenOS's core that determines how software is supposed to be built, and this change merely reflects that simple fact.

vhotspur:
> > I think CI script does the "build everything" part a bit better
> 
> It does, but running CI on local changes is somewhat laborious. My intention is to allow the "build everything" without any setup at all (analogous in simplicity to `make check`), and I think this does the job well. I break ports often enough to appreciate the simplicity.

Can you elaborate a bit more on _somewhat laborious_? The command requires two switches and it can build _everything_ or a selected subset. In parallel, with in-VM tests and with artefacts for manual experimenting.

By the way, right now the building of everything takes few hours (and the QEMU branch took about 3 hours for one architecture). It does not feel right to me to pretend this is something you should run on each commit.

> > I think this pollutes the HelenOS tree more than it deserves but if you believe it helps you in streamlining libposix then go ahead :-).
> 
> I do believe that. :-) Out of curiosity, how would you, in this regard, compare current harbour, tools/xcw, and this change? IMO current harbour pollutes mainline by its very nature as "something that breaks if you change mainline, and it's our problem to keep it fixed". In this sense, the entanglement is already at maximum.

It breaks only if you make changes in one (rather small) library that is not the flagship of HelenOS but rather a glue layer.

> > I think Git submodules are bad in general and they would eventually bite us.
> 
> Can you elaborate on this? I consider them a very helpful tool, so I'd like to hear why you think them troublesome.

Being no expert in Git what I found most troublesome is that they `git status` somewhat hides changes in the submodule and submodule update can sometimes even overwrite them.


> With ports, security is not a concern (if one repo is compromised, both are), but at this moment, the repositories are nowhere close to being independent. Yes, it's the ideal we should strive for, but it's not currently consistent with how things are.
> 
> Oh, also, it might not be clear from this pull request, but the intention is for the scripts in the harbours repo to be removed. Otherwise this would just add to the maintanance burden.

The separation has also the advantage that you can test different ports with different branches of HelenOS which I would consider as independence. Your proposal binds them together. By the way, I would really expect such big change to be mentioned in the PR description ;-)

If you are planning to remove the scripts than move the package HARBOURs too. Or, do not use a submodule but rather clone the repository as JS suggested to allow cloning from a different fork.

> Basically, the problem I'm solving is that current hsct.sh digs into internal details of the build scripts to figure out which flags to use, and then plays a guessing game about which flags to keep and which to filter out. That's one of the worst possible designs. It's HelenOS's core that determines how software is supposed to be built, and this change merely reflects that simple fact.

While XCW just copied them. We can do the same in Coastline if you think that is better. And by the way, it is not a guessing game, it simply overrides include paths and removes warning flags. Because some options already needed to be taken from the Makefiles it was straightforward to get the flags too.

le-jzr:
> Can you elaborate a bit more on somewhat laborious?

You need to separately maintain three independent directories (ci, mainline, harbours), make sure the correct revisions of each are checked out, make sure your local changes are commited (i.e. no test-building uncommited changes), and switch directories back and forth to run the desired scripts. Further, you can't build non-default configurations (or can you? I haven't really tried tbh), and the scripts are so fragile that almost any change that actually needs to be tested with ports needs updates to the scripts. Case in point, current breakage.

> It breaks only if you make changes in one (rather small) library that is not the flagship of HelenOS but rather a glue layer.

That's false. Any change to libc or the build process affects ports. Just recently, some broke simply because I added the `intmax_t` type. Overall, the whole setup is incredibly fragile.

> The separation has also the advantage that you can test different ports with different branches of HelenOS which I would consider as independence.

I cannot think of a single scenario where the separation makes things easier. You can test different ports with different branches just fine with submodules. If anything breaks with submodules, it means it would have broken with current arrangement even worse.

> Or, do not use a submodule but rather clone the repository as JS suggested to allow cloning from a different fork.

Uh, how does that "allow" anything you can't do with a submodule? If anything, submodule makes this easier.

le-jzr:
> By the way, right now the building of everything takes few hours (and the QEMU branch took about 3 hours for one architecture). It does not feel right to me to pretend this is something you should run on each commit.

On my PC, building a single configuration (which accounts for most possible issues) takes 7 minutes, including all downloads. I choose to pretend that's okay to run on a casual basis.

le-jzr:
@vhotspur Tbh I'm a little stuck. For some reason I'm suddenly unable to get even basic harbour build to work locally (fails with `[hsct]: Error: Unsupported architecture: $(UARCH) = ''.`), and I have no idea what caused it. I thought I would just modify CI to work with this change, but after looking at the code, it seems to me like just understanding what's going on in it would be an excessive amount of work.

vhotspur:
> > Can you elaborate a bit more on somewhat laborious?
>
> You need to separately maintain three independent directories (ci, mainline, harbours), make sure the correct revisions of each are checked out, make sure your local changes are commited (i.e. no test-building uncommited changes), and switch directories back and forth to run the desired scripts.

Wrong. CI scripts support building uncommitted changes via `wip://` prefix (see [this wiki page](http://www.helenos.org/wiki/CI)). If not specified otherwise, they clone upstream.

> Further, you can't build non-default configurations (or can you? I haven't really tried tbh), and the scripts are so fragile that almost any change that actually needs to be tested with ports needs updates to the scripts. Case in point, current breakage.

Not now. I knew it would be nice to test with LLVM compiler too but I have not found enough time to implement it. Regarding generic other options, how would you like to specify them on the command-line (I have not thought about this yet)?

> > It breaks only if you make changes in one (rather small) library that is not the flagship of HelenOS but rather a glue layer.
> 
> That's false. Any change to libc or the build process affects ports. Just recently, some broke simply because I added the `intmax_t` type. Overall, the whole setup is incredibly fragile.

You are mixing two things here. The multi-repository setup is not fragile. It corresponds with logical separation between our code (i.e. written from scratch), ported code and tooling for nightly builds. Changing build process is a non-trivial change that is bound to affect many things and obviously it affects tools that are calling it. What is fragile is `libposix` and its dependency on `libc`. What broke are the `configure` scripts that are like mad trying to guess what system they are compiling for and supply their own implementations when none could be found. So, yeah, adding `intmax_t` have broken many packages because now the `configure` is taking a different path that was not taken before. That has nothing to do with how things are built.

A sidenote: I am well aware of this fragility of `libposix` and that is why I have started playing with `libinux` as a possible alternative. But again, this is not a problem of the Coastline scripts but of the incompleteness of the POSIX layer. Not talking about the fact that most applications actually expects Linux underneath and I seriously doubt they would compile elsewhere without any (no matter how tiny) modifications.

For the record, I am not saying that `hsct.sh` is the best script I have ever written. But it works well given the conditions. I am planning to replace it with something Pythonic to improve readability and maintainability but I touched the rewrite some 3 months ago and have not any time since that.

> How does that "allow" anything you can't do with a submodule? If anything, submodule makes this easier.

I will take your word for it. My opinion is that submodule is not the right thing here and I will leave it there.
   
> > By the way, right now the building of everything takes few hours (and the QEMU branch took about 3 hours for one architecture). It does not feel right to me to pretend this is something you should run on each commit.
> 
> On my PC, building all ports in a single configuration (which accounts for most possible issues) takes 7 minutes, including all downloads.

I would like to have that Cray machine at home too :-). Looking at [this build](http://ci.helenos.org/build-112/index.html#harbour-gcc) (will disappear probably soon) only GCC alone builds for more than 7 minutes on my machine.


>   @vhotspur Tbh I'm a little stuck. For some reason I'm suddenly unable to get even basic (vanilla) harbour build to work locally (fails with `[hsct]: Error: Unsupported architecture: $(UARCH) = ''.`), and I have no idea what caused it. I thought I would just modify CI to work with the change I'm proposing, but after looking at the code, it seems to me like just understanding what's going on in it would be an excessive amount of work.

CI just calls `hsct.sh` so you have to fix it there. This looks like it is unable to fetch the `UARCH` variable from some `Makefile` to properly set-up cross-targets. I will try to look at this soon but I cannot promise anything.

le-jzr:
> Wrong. CI scripts support building uncommitted changes via wip:// prefix

Ah, I somehow missed that. Thanks for pointing that out.

> Regarding generic other options, how would you like to specify them on the command-line (I have not thought about this yet)?

You don't have to. You can just configure using existing UI and build using that configuration.

> It corresponds with logical separation between our code (i.e. written from scratch), ported code and tooling for nightly builds.

IMO this separation is an illusion and doesn't really exist in practice. I don't consider it a well-defined abstraction boundary.

> I would like to have that Cray machine at home too :-)

Just Ryzen 7 with loads of RAM. :-)

le-jzr:
In response to the feedback, I decided I won't be merging this pull request in its current form. I'll open a more conservative proposal shortly.

PR #12, opened 2018-01-11
Use Travis CI for per-commit builds

vhotspur:
Any objections of having `.travis.yml` with [Travis CI](https://travis-ci.org/) configuration? The configuration is a very basic one that only builds for all architectures. Unlike [our CI](http://ci.helenos.org) Travis does not provide any downloadable artifacts but Travis can be run on forks and also on pull requests.

We might add building of ports later but I am not sure if we would be able to fit into the 2h limit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,35 @@
+language: c
+
+env:
+  global:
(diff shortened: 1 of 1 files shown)


le-jzr:
LGTM :)

PR #14, opened 2018-01-12
async_usleep: no need to use a heap-allocated amsg_t

Aearsis:
I somehow feel this must be based on wrong assumptions, because otherwise it wouldn't be done this way. Feel free to reject if so.


le-jzr:
Looks good to me. :)

PR #15, opened 2018-01-16
Generate files that a third party can use to build software for HelenOS

le-jzr:
without necessarily knowing anything about the build system.

Not yet ideal in terms of forward compatibility, so I expect some small changes will be necessary, but I'm fairly comfortable with this level of abstraction. The main repository doesn't need to know anything about ports, and ports don't need to know anything particular about HelenOS (except where to look for the configuration).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ CONFIG_HEADER = config.h
 ERRNO_HEADER = abi/include/abi/errno.h
 ERRNO_INPUT = abi/include/abi/errno.in
 
-.PHONY: all precheck cscope cscope_parts autotool config_auto config_default config distclean clean check releasefile release common boot kernel uspace
+.PHONY: all precheck cscope cscope_parts autotool config_auto config_default config distclean clean check releasefile release common boot kernel uspace export-posix
 
 all: kernel uspace
 	$(MAKE) -r -C boot PRECHECK=$(PRECHECK)
@@ -59,6 +59,13 @@ kernel: common
 
 uspace: common
 	$(MAKE) -r -C uspace PRECHECK=$(PRECHECK)
+
+export-posix: common
+ifndef EXPORT_DIR
+	@echo ERROR: Variable EXPORT_DIR is not defined. && false
+else
+	$(MAKE) -r -C uspace export EXPORT_DIR=$(abspath $(EXPORT_DIR))
+endif
 
 precheck: clean
 	$(MAKE) -r all PRECHECK=y
--- a/tools/autotool.py
+++ b/tools/autotool.py
@@ -614,6 +614,7 @@ def main():
 					print_error(["Toolchain for target is not installed, or CROSS_PREFIX is not set correctly."])
 				path = "%s/%s/bin" % (cross_prefix, platform)
 		
+		common['TARGET'] = target
 		prefix = "%s-" % target
 		
 		# Compiler
--- a/uspace/Makefile
+++ b/uspace/Makefile
@@ -254,12 +254,15 @@ BUILDS_TESTS := $(addsuffix .build-test,$(DIRS) $(LIBS) $(BASE_LIBS))
 DEPS = $(addsuffix /deps.mk,$(DIRS) $(LIBS))
 CLEANS := $(addsuffix .clean,$(DIRS) $(LIBS) $(BASE_LIBS))
 
-.PHONY: all $(BASE_BUILDS) $(BUILDS) $(BUILDS_TESTS) $(CLEANS) clean
+.PHONY: all $(BASE_BUILDS) $(BUILDS) $(BUILDS_TESTS) $(CLEANS) clean export
 
 all: $(BUILDS) $(BUILDS_TESTS)
 
 $(BUILDS_TESTS): $(BASE_BUILDS) $(BUILDS)
 	$(MAKE) -r -C $(basename $@) all-test PRECHECK=$(PRECHECK)
+
+export: lib/posix.build lib/math.build lib/clui.build
+	$(MAKE) -r -C lib/posix export EXPORT_DIR=$(EXPORT_DIR)
 
 clean: $(CLEANS)
 	find lib app drv srv -name '*.o' -follow -exec rm \{\} \;
--- a/uspace/Makefile.common
+++ b/uspace/Makefile.common
@@ -173,31 +173,57 @@ endif
 	TEST_BINARY_LIBS += $(TEST_LIBS)
 endif
 
-COMMON_CFLAGS = $(INCLUDES_FLAGS) -O$(OPTIMIZATION) -imacros $(CONFIG_HEADER) \
-	-ffreestanding -nostdlib -nostdinc -fexec-charset=UTF-8 \
-	-finput-charset=UTF-8 -D__$(ENDIANESS)__ -D_HELENOS_SOURCE -fno-common \
-	-fdebug-prefix-map=$(realpath $(ROOT_PATH))=.
-
-GCC_CFLAGS = -ffunction-sections -Wall -Wextra -Wno-clobbered \
-	-Wno-unused-parameter -Wmissing-prototypes -std=gnu99 \
-	-Werror-implicit-function-declaration \
-	-Wwrite-strings -pipe
-
-# -Wno-missing-prototypes is there because it warns about main().
-# This should be fixed elsewhere.
-CLANG_CFLAGS = -Wall -Wextra -Wno-unused-parameter -Wno-missing-field-initializers -Wno-typedef-redefinition \
-	-Wno-missing-prototypes -Wno-unused-command-line-argument \
-	-std=gnu99 -Werror-implicit-function-declaration -Wwrite-strings \
-	-pipe -fno-stack-protector -fno-PIC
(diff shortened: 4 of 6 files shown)


vhotspur:
I very much like the overall idea as it can greatly simplify things.

But I would prefer something more self-contained. My idea when we discussed this during the telco was that the includes and libraries would be copied to a separate folder with the following layout:

```text
helenos/helenos.specs
helenos/include/ # all headers of libposix (libc and others in subdirs)
helenos/libs/ # all the needed .a files
helenos/config.mk # vars for make
helenos/config.sh # vars for shell
```

and that Coastline would copy this whole directory and use it. And I expected that the specs file would contain all the needed switches so one would be able to specify the `CFLAGS` (`LDFLAGS`, ...) as `-specs helenos.specs` and `--helenos-root=<path to the copied dir>` and the specs file would append all the `-Wl,--whole-archive` etc.

Or perhaps the specs file would also understand switches `--helenos-native` and `--helenos-posix` that would enable a particular set of `-I` and `-L`. I believe such setup would be useful for XCW too.

(Thus the `config.*` files would contain paths to the compiler, architecture etc. but all compilation-related stuff would be hidden inside the specs file.)



What is the expected usage scenario if someone decides to create a better Coastline? How is the export supposed to be used?

Thanks!

le-jzr:
I like the idea of a self-contained folder that could be copied as a unit. That would certainly be simpler than what it's currently doing.

On the other hand, I would prefer not making it dependent on specs file, since that is exclusively a GCC feature, and would prevent using coastline with clang or any other compiler. It also assumes that specs file is always used, and prevents further simplifying things with a true HelenOS cross-compiler. For similar reasons, I think adding custom switches would be harmful and ultimately pointless. The use case of having both "native" and "posix" configurations available is easily solved by providing two sets of variables.

As for the usage scenario, it's mostly described in comments in `uspace/Makefile`. The user runs `make build_vars`, reads/includes `uspace/build_vars.mk` (actually something I want to change, since this is still more knowledge of the directory layout than I like), copies the libraries and associated files to a directory of their choosing, possibly headers as well, adds `$(HELENOS_CROSS_PATH)` to `PATH`, and uses target-prefixed executables and provided driver flags for compilation. The one major assumption present is that assembly/linking is always done using the compiler driver (as it should be), which in some cases might require adjustments to the third party build scripts, but at worst that can be resolved by adding more variables.

le-jzr:
What if instead of providing a complete directory for the user to copy, the user instead provided a path and the Makefile would copy files into it directly? E.g. the user would run `make build_vars BUILD_DIR="path/to/build/dir"` which would put all the necessary files there. Names subject to bikeshedding, of course.

vhotspur:
> On the other hand, I would prefer not making it dependent on specs file, since that is exclusively a GCC feature, and would prevent using coastline with clang or any other compiler. It also assumes that specs file is always used, and prevents further simplifying things with a true HelenOS cross-compiler. For similar reasons, I think adding custom switches would be harmful and ultimately pointless. The use case of having both "native" and "posix" configurations available is easily solved by providing two sets of variables.

We can still generate the specs file for GCC and have the `CFLAGS` variable if someone uses a different compiler. The specs file has the big advantage of making the command-line parsable by a human. Another option would be to generate a wrapper shell script that looks like GCC and provide our own `bin` folder. But as long as this compiles the packages I really do not care that much.

One more thing. Would it be possible to provide a separate variable for **preprocessor flags**? We use `CPPFLAGS` in several packages where `$(CC) -E` is invoked directly during build.

> As for the usage scenario, it's mostly described in comments in `uspace/Makefile`.

Cool :-).

> The one major assumption present is that assembly/linking is always done using the compiler driver (as it should be), which in some cases might require adjustments to the third party build scripts, but at worst that can be resolved by adding more variables.

Does this play well with `libtool`? I was unable to try it so that is why I am asking...

> What if instead of providing a complete directory for the user to copy, the user instead provided a path and the Makefile would copy files into it directly? E.g. the user would run `make build_vars BUILD_DIR="path/to/build/dir"` which would put all the necessary files there. Names subject to bikeshedding, of course.

It would be nice if the export would not contain any absolute file paths. But I am not sure if that is possible at all.

le-jzr:
> One more thing. Would it be possible to provide a separate variable for __preprocessor flags__? We use `CPPFLAGS` in several packages where `$(CC) -E` is invoked directly during build.

I wonder, are there any cases where we can't use CFLAGS for this? gcc just ignores any valid flags that are not relevant for its current run. But I don't mind splitting them up.

>  Does this play well with libtool? I was unable to try it so that is why I am asking...

It works for packages in harbours on my system. That's about as much as I can say with certainty. :)

> It would be nice if the export would not contain any absolute file paths. But I am not sure if that is possible at all.

I don't think it contains any, except for the one path derived from `CROSS_PREFIX`. I'm not 100% sure, but I don't see why any other absolute paths should be there.

le-jzr:
> One more thing. Would it be possible to provide a separate variable for __preprocessor flags__? We use `CPPFLAGS` in several packages where `$(CC) -E` is invoked directly during build.

I wonder, are there any cases where we can't use CFLAGS for this? gcc just ignores any valid flags that are not relevant for its current run. But I don't mind splitting them up.

>  Does this play well with libtool? I was unable to try it so that is why I am asking...

It works for packages in harbours on my system. That's about as much as I can say with certainty. :)

> It would be nice if the export would not contain any absolute file paths. But I am not sure if that is possible at all.

I don't think it contains any, except for the one path derived from `CROSS_PREFIX`. I'm not 100% sure, but I don't see why any other absolute paths should be there.

vhotspur:
> > One more thing. Would it be possible to provide a separate variable for __preprocessor flags__? We use `CPPFLAGS` in several packages where `$(CC) -E` is invoked directly during build.
> 
> I wonder, are there any cases where we can't use CFLAGS for this? gcc just ignores any valid flags that are not relevant for its current run. But I don't mind splitting them up.

I do not have a MWE but I remember that there were some issues when `CPPFLAGS` were set to `CFLAGS` when porting GCC.

le-jzr:
New version. I did as I suggested earlier and "install" the files into a path provided by the caller. I think its cleaner this way, and requires a lot less work by the consumer.

vhotspur:
I like this. I have some comments to the Coastline-side but this looks okay to me.

To make the export a bit more movable, would you consider the following patch? It needs to split the `config` file into two, but it is actually possible to move the generated directory around.

Up to you, otherwise it is *go ahead with the merge here* from me.

```diff
diff --git a/uspace/lib/posix/Makefile b/uspace/lib/posix/Makefile
index 6781ed0b3..037ade270 100644
--- a/uspace/lib/posix/Makefile
+++ b/uspace/lib/posix/Makefile
@@ -97,11 +97,11 @@ TEST_SOURCES = \
 	test/scanf.c
 
 EXPORT_CPPFLAGS = \
-	-specs $(EXPORT_DIR)/lib/gcc.specs \
-	-isystem $(EXPORT_DIR)/include
+	-specs $$(HELENOS_EXPORT_ROOT)/lib/gcc.specs \
+	-isystem $$(HELENOS_EXPORT_ROOT)/include
 
 EXPORT_LDFLAGS = \
-	-L$(EXPORT_DIR)/lib \
+	-L$$(HELENOS_EXPORT_ROOT)/lib \
 	--whole-archive -lc -lmath --no-whole-archive \
 	-T link.ld
 
@@ -130,10 +130,11 @@ $(INCLUDE_LIBC): $(shell find ../c/include -name '*.h')
 	cp -r -L --remove-destination -T ../c/include $@
 	find ../c/include -type f -and -not -name '*.h' -delete
 
-export: $(EXPORT_DIR)/config
+export: $(EXPORT_DIR)/config.mak $(EXPORT_DIR)/config.rc
 
-$(EXPORT_DIR)/config: export-libs export-includes
+$(EXPORT_DIR)/config.mak: export-libs export-includes
 	echo '# Generated file, do not modify.' >> $@.new
+	echo '# Do not forget to set HELENOS_EXPORT_ROOT.' >> $@.new
 	echo 'HELENOS_CROSS_PATH="$(shell dirname $(CC))"' >> $@.new
 	echo 'HELENOS_TARGET="$(TARGET)"' >> $@.new
 	echo 'HELENOS_CPPFLAGS="$(EXPORT_CPPFLAGS)"' >> $@.new
@@ -141,6 +142,10 @@ $(EXPORT_DIR)/config: export-libs export-includes
 	echo 'HELENOS_LDFLAGS="$(EXPORT_LDFLAGS)"' >> $@.new
 	mv $@.new $@
 
+$(EXPORT_DIR)/config.rc: $(EXPORT_DIR)/config.mak
+	sed 's:$$(HELENOS_EXPORT_ROOT):$$HELENOS_EXPORT_ROOT:g' < $< >$@
+
+
 export-libs: $(EXPORT_FILES) export-includes
 	mkdir -p $(EXPORT_DIR)/lib
 	cp -L $(EXPORT_FILES) $(EXPORT_DIR)/lib
```

vhotspur:
> Isn't `.mk` a more common extension for Make files though?

I have seen `.mk`, `.mak` and even `.make`. Up to you.

le-jzr:
I'll go with `.mk`, then.  I like it more. :)

PR #16, opened 2018-01-24
Travis CI improvements

vhotspur:
This PR contains several improvements of the CI process. The dashboard is less cluttered as less variables are set (see [here](https://travis-ci.org/vhotspur/helenos/builds/332773178)). Ported software is built too. But only for `ia32` as that seems to be the most healthy platform in this context. Overall, this should provide shorter feedback than the [nightly builds](http://ci.helenos.org).

Due to complexity of the building process (especially when ported software is compiled) a separate script is used with Travis CI to control the build. See comments in `.travis.yml` for details how to customize what is being built (if you plan to use it on your feature branches).

Okay to merge this?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,35 +1,62 @@
-language: c
+#
+# Travis CI configuration notes
+# -----------------------------
+#
+# * HelenOS is always built for all architectures
+#   * that is done using matrix environmental variables
+#   * the list of architectures has to be kept manually in sync with
+#     HelenOS.config
+#   * the actual architecture is specified via H_ARCH variable
+#
+# * ported software is built when H_HARBOURS is set to true for given
+#   configuration
+#   * building ported software takes a lot of time, it is thus impossible
+#     to build for all architectures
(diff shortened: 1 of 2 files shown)


le-jzr:
I don't see why not.

le-jzr:
Actually, on second thought...
I don't like that it makes the CI depend on a latest revision of a separate repository. IMO CI runs shouldn't be affected by commits to an independent repo.

vhotspur:
> I don't like that it makes the CI depend on a latest revision of a separate repository. IMO CI runs shouldn't be affected by commits to an independent repo.

I understand your concerns (though I am still not convinced about the solution you proposed). What if we add Travis to the Harbours repo too? That way the breakage would appear in two places and at least in one of the cases the developer that was actually responsible would get the e-mail...

The main point of this new script is to provide shorter feedback that the nightly builds.

le-jzr:
> What if we add Travis to the Harbours repo too?

That would help. :)

PR #17, opened 2018-01-29
libdrv: fix leaking ddf_dev_t reference

Aearsis:
Even though the permanent reference held by the device being added is
dropped correctly, the temporary reference added by remove/gone is not.
That caused leaking devices and parent sessions along with them (which
is the way we discovered it).


Aearsis:
It seems that devices were never supposed to be deleted :) Now the device still remains in the list, and second removal pagefaults. I'll continue debugging...

jermar:
Good job!

PR #22, opened 2018-02-28
Remove whitespace on empty lines

le-jzr:
I'm well aware that some people will dislike this, and claim we should have
indentation on empty lines. My counterpoint is that either way, this needs
to be consistent, and in current code, whitespace on empty lines is very
inconsistent. Removing whitespace is very simple to do automatically,
thus ensuring consistency. A claim could be made that we could as well
create automated script for automatic indentation, and use that instead.
However, if you have that script already at hand, you can just as well work
with the indented style by having your editor apply the conversions
automatically.

Thoughts?


jermar:
I'd say just do it. Also no need for such a defensive commit message.

le-jzr:
Right, I reworded it.

PR #23, opened 2018-03-06
[unfinished] A replacement for autogen.py

le-jzr:
I haven't finished updating all architectures (it's a chore), but thought I might put it up for review early.

General idea: instead of a script generating both C header and asm offsets from the verbose description we have, write a valid C header directly and generate just the offsets.


jermar:
It's a chore, because it took quite a lot of effort to get to this point. What benefit will this change bring and why is it necessary?

le-jzr:
It's a chore because I'm copying over the relevant comments from *.ag to the new headers, which means going one file at a time instead of just using a script. ;)

It's a simplification. The benefit is that the type definition in a regular header is easier to read and use. There's also the simplification of dependencies. Previously, some headers would need other headers to be generated first, which adds more stages to the process. Now the headers all exist a priori, so less work is needed to make them build correctly. In addition, in some cases the offsets aren't even used (all of arm32, for example), so there is fewer generated headers. This has little influence on the current build scripts of course, but it makes it easier to switch to another build system, which is why I spent time on this in the first place.

le-jzr:
I don't feel like this is the best possible solution, so closing for now.

PR #24, opened 2018-03-08
Turn context_save/context_restore into standard setjmp/longjmp.

le-jzr:
Since they already _almost_ are identical, this just closes the tiny gap and removes an ugly hack from the standard library in the process.

The assembly changes need to be reviewed by more eyes though.

### DIFF
--- a/uspace/lib/c/arch/amd64/src/fibril.S
+++ b/uspace/lib/c/arch/amd64/src/fibril.S
@@ -34,9 +34,9 @@
 ## Save current CPU context
 #
 # Save CPU context to context_t variable
-# pointed by the 1st argument. Returns 0 in EAX.
+# pointed by the 1st argument. Returns 0 in RAX.
 #
-FUNCTION_BEGIN(setjmp)
+FUNCTION_BEGIN(__setjmp)
 	movq (%rsp), %rdx     # the caller's return %eip
 
 	# in %rdi is passed 1st argument
@@ -53,14 +53,14 @@ FUNCTION_BEGIN(setjmp)
 	movq %fs:0, %rax
 	movq %rax, CONTEXT_OFFSET_TLS(%rdi)
 
-	xorq %rax, %rax                      # setjmp returns 0
+	xorq %rax, %rax                      # __setjmp returns 0
 	ret
-FUNCTION_END(setjmp)
+FUNCTION_END(__setjmp)
 
 ## Restore current CPU context
 #
 # Restore CPU context from context_t variable
-# pointed by the 1st argument. Returns RSI in EAX.
+# pointed by the 1st argument. Returns second argument in RAX.
 #
 FUNCTION_BEGIN(__longjmp)
 	movq CONTEXT_OFFSET_R15(%rdi), %r15
--- a/uspace/lib/c/arch/amd64/src/fibril.S
+++ b/uspace/lib/c/arch/amd64/src/fibril.S
@@ -34,9 +34,9 @@
 ## Save current CPU context
 #
 # Save CPU context to context_t variable
-# pointed by the 1st argument. Returns 1 in EAX.
+# pointed by the 1st argument. Returns 0 in EAX.
 #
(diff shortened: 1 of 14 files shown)


uspace/lib/c/arch/ppc32/src/fibril.S
```diff
@@ -89,13 +89,13 @@ FUNCTION_BEGIN(context_restore)
...
 
-	lwz r4, CONTEXT_OFFSET_PC(r3)
-	mtlr r4
+	lwz r5, CONTEXT_OFFSET_PC(r3)
+	mtlr r5
 
-	# context_restore returns 0
-	li r3, 0
+	# __longjmp returns second argument
+	li r3, r4
```
le-jzr: This is probably wrong. `li` stands for load immediate?

uspace/lib/c/arch/ia64/include/libarch/fibril.h
```diff
@@ -42,7 +42,7 @@
 #include <libarch/fibril_context.h>
 
 /*
- * context_save() and context_restore() are both leaf procedures.
+ * setjmp() and __longjmp() are both leaf procedures.
```
jermar: The standard says setjmp is a macro.

uspace/lib/c/arch/ppc32/src/fibril.S
jermar: Yeah, very likely. Can't you just mov or move? In the worst case, RTFM :-)

uspace/lib/c/arch/ia64/include/libarch/fibril.h
le-jzr: Ah, indeed, I forgot about that. `#define setjmp(env) setjmp(env)` should do the trick.

uspace/lib/c/arch/amd64/src/fibril.S
```diff
@@ -53,17 +53,16 @@ FUNCTION_BEGIN(context_save)
...
+	xorq %rax, %rax                      # setjmp returns 0
 	ret
-FUNCTION_END(context_save)
+FUNCTION_END(setjmp)
 
 ## Restore current CPU context
 #
 # Restore CPU context from context_t variable
-# pointed by the 1st argument. Returns 0 in EAX.
+# pointed by the 1st argument. Returns RSI in EAX.
```
jermar: Maybe rather  "Returns second argument in EAX"

le-jzr:
@jermar, I tested it on all emulated platforms except for IA64 (I wasn't able to build ski), which coincidentally is the platform I'm least sure about being correct. :)

jermar:
The ia64 changes look good. Ski can often be installed as a package in your favorite Linux distribution.

jxsvoboda:
I don't understand why you are renaming it to __setjmp()/__longjmp(). Either they comply with the standard and you would name them setjmp()/longjmp(), or they don't , in which they are not setjmp and longjmp. Why can't they use HelenOS naming conventions then? What's wrong with the names context_save and context_restore then?

le-jzr:
That's a good question. Thinking back, there's a history to that. Originally, I was forgetting that `setjmp()` is defined as a macro, so I thought it needs to be a function symbol. That would make it impossible to implement `setjmp()` by wrapping `context_save()` in C, so I renamed `context_save()` to `setjmp()`, so as to avoid more assembly. Normally, I'd change `context_restore()` to `longjmp()`, but that would require an extra conditional branch in assembly for all architectures. Ick. So I made it `__longjmp()`. Then @jermar pointed out that `setjmp()` is only defined as a macro, so I changed the name to `__setjmp()` for consistency. It also expresses that code calling `__setjmp()/__longjmp()` instead of `setjmp()/longjmp()` depends on properties not guaranteed by the standard. But at this point, the motivation for going with the `setjmp`/`longjmp` naming scheme is just cached historical outcome, which I failed to reevaluate.

So that's for the explanation of why. I'll rename them back if you wish.

PR #25, opened 2018-03-13
Noreclaimers

jermar:
This branch contains the removal of the so-called capability reclaiming which simplifies capability allocation in that it is no longer necessary to go through the existing capabilities in an attempt to reclaim some unused one. In the course of these changes, I slightly refactored the ipc_cleanup() mechanism and provided a race-free implementation of connect_me_to and connect_to_me. In the new scheme, the new phone is first connected and only then is its capability published (made visible to userspace). Thus it cannot be fiddled with from userspace in the meantime and pathological ipc_cleanup() scenarios such as connected phones that are only reachable from the answerbox they are connected to (but not from their own task) cannot occur.

### DIFF
--- a/kernel/generic/include/cap/cap.h
+++ b/kernel/generic/include/cap/cap.h
@@ -63,9 +63,7 @@ struct call;
 struct irq;
 struct phone;
 
-struct kobject;
 typedef struct kobject_ops {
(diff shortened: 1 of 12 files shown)


jermar:
Interesting, this one did not get closed automatically.

PR #26, opened 2018-03-16
Clean up PAGE_* flags.

le-jzr:
Remove "nop flags", they are confusing for readers and any benefit they would
gain in self-documentation they lose in being used inconsistently.

Remove "PAGE_READ", the only place it's used meaningfully is the incomplete
RISC-V implementation, and most call sites assume read is implied.


jermar:
I'd rather like to see PAGE_NOT_CACHEABLE or PAGE_KERNEL or PAGE_READ in the flags instead of mere 0. Also, if we redefine these not to be 0 one day (for whatever reason, e.g. architecture that supports execute only mode), they are already in place and we should rather fix the places where they are now silently implied (if that is the inconsistency you are referring to).

le-jzr:
Then I would propose an alternative:
  - Flags must contain exactly one of `PAGE_READ_ONLY`, `PAGE_READ_WRITE`, `PAGE_EXECUTE_ONLY` (`PAGE_EXECUTE` stays for combination with the former three), checked by an `assert()`. I omitted a write-only and write-execute mode because I don't think that would be meaningful.
 - Have PAGE_NOT_CACHEABLE, but make pages cacheable by default, as it's arguably the "normal" choice. This way, `flags | PAGE_NOT_CACHEABLE` works as expected.

jermar:
I am for mutually exclusive PAGE_CACHEABLE / PAGE_NOT_CACHEABLE and PAGE_USER / PAGE_KERNEL.

In case of page protections, some combinations don't make sense, as you pointed out. The valid ones can be conveniently described by a set of enum values. The disadvantage is that it would be more difficult to test for individual protections (e.g. among others, is this page writable?). Otherwise I'd prefer to have a non-zero bit for each protection. For the record, ia64 makes it possible to have execute only pages.

PR #27, opened 2018-03-18
Implement __multi3 by renaming __muldi3

jermar:
GCC [documentation](https://gcc.gnu.org/onlinedocs/gccint/Integer-library-routines.html) defines __muldi3 and __multi3 as follows:

> Runtime Function: long __muldi3 (long a, long b)
> Runtime Function: long long __multi3 (long long a, long long b)
> 
>     These functions return the product of a and b.

Interestingly our libsoftint already contains a function that operates and returns long long, but is wrongly called __muldi3. This PR renames it to __multi3, provides a new trivial implementation of __muldi3 via __multi3 and adds a set of PCUT tests.

Please take a look at these changes. Note that I am currently looking into solving the FIXME in [strtol.c:59](https://github.com/HelenOS/helenos/blob/973be38782462a672111ff6833f8dbb40d2a44bf/uspace/lib/c/generic/strtol.c#L59), but am hitting some issues there.


le-jzr:
Actually, it's the GCC documentation that is incorrect. Instead, check the implementation of those functions in libgcc or LLVM's compiler-rt. If I remember correctly, `di` suffix is always used for 64b arguments, and `ti` suffix for 128b arguments. So `__multi3` is actually a 128b multiplication.

I fell into the same trap recently, see commits https://github.com/HelenOS/helenos/commit/83488467de531542d2952d3e89d16ece20eaa46d, https://github.com/HelenOS/helenos/commit/aec41c815c786936ee5edb382011403392833e23, https://github.com/HelenOS/helenos/commit/2467b4174c7fecef314011212ab09a004872362d, https://github.com/HelenOS/helenos/commit/17c1427397e6ec84fdd0ce49eddfd8c5ffb03b62.

le-jzr:
I could swear I remember putting some comments about it in libsoftint. I must have forgot to push them. -_-

jermar:
Heh, thanks for the correction!

PR #28, opened 2018-03-21
Change the way page flags work.

le-jzr:
This is the first part of a larger overhaul of the generic paging interface.

Page flags in `generic/mm/page.h` are expanded to include a "discriminant" -- a single enum-like flag that determines the contents of the page table entry. This is currently one of "not present", "next level page table", "leaf page". Leaf pages further have a separate discriminant for each legal combination of read-write-execute permissions.

In addition to discriminant, there are two mutually-exclusive pairs of flags: "cacheable"/"not-cacheable", and "kernel"/"user". Leaf pages must have one flag from each pair. Non-leaf entries may not have cacheability flag (the meaning of cacheability on non-leaf entries may not be consistent between platforms, and the generic code doesn't know the appropriate flags anyway). Non-leaf entries may have one of "kernel"/"user" flags, but are not required to. The flag is defined to mean that everything in the subtree is kernel-only or user-only, respectively. The motivation for this feature is that some platforms can disable kernel access to user pages, which is useful as a safety and/or debugging precaution.

### DIFF
--- a/kernel/genarch/src/mm/page_ht.c
+++ b/kernel/genarch/src/mm/page_ht.c
@@ -60,7 +60,7 @@ static void ht_remove_callback(ht_link_t *);
 static void ht_mapping_insert(as_t *, uintptr_t, uintptr_t, unsigned int);
 static void ht_mapping_remove(as_t *, uintptr_t);
 static bool ht_mapping_find(as_t *, uintptr_t, bool, pte_t *);
-static void ht_mapping_update(as_t *, uintptr_t, bool, pte_t *);
+static void ht_mapping_update(as_t *, uintptr_t, bool, const pte_t *);
 static void ht_mapping_make_global(uintptr_t, size_t);
 
 slab_cache_t *pte_cache = NULL;
@@ -167,8 +167,8 @@ void ht_mapping_insert(as_t *as, uintptr_t page, uintptr_t frame,
 		assert(pte != NULL);
 
 		pte->g = (flags & PAGE_GLOBAL) != 0;
-		pte->x = (flags & PAGE_EXEC) != 0;
-		pte->w = (flags & PAGE_WRITE) != 0;
+		pte->x = (flags & _PAGE_EXEC) != 0;
+		pte->w = (flags & _PAGE_WRITE) != 0;
 		pte->k = !(flags & PAGE_USER);
 		pte->c = (flags & PAGE_CACHEABLE) != 0;
 		pte->p = !(flags & PAGE_NOT_PRESENT);
@@ -267,7 +267,7 @@ bool ht_mapping_find(as_t *as, uintptr_t page, bool nolock, pte_t *pte)
  * @param nolock   True if the page tables need not be locked.
  * @param pte      New PTE.
  */
-void ht_mapping_update(as_t *as, uintptr_t page, bool nolock, pte_t *pte)
+void ht_mapping_update(as_t *as, uintptr_t page, bool nolock, const pte_t *pte)
 {
 	irq_spinlock_lock(&page_ht_lock, true);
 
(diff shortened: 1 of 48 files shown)


jermar:
I skimmed through the changes once today, will have a more thorough look probably during the weekend. On what platforms have you tested this?

kernel/generic/include/mm/mm.h
```diff
@@ -35,31 +35,116 @@
...
-#define PAGE_PRESENT_SHIFT		1
-#define PAGE_NOT_PRESENT_SHIFT		PAGE_PRESENT_SHIFT
-#define PAGE_USER_SHIFT			2
-#define PAGE_KERNEL_SHIFT		PAGE_USER_SHIFT
+#include <assert.h>
+#include <stdbool.h>
+
+#define PAGE_NOT_CACHEABLE_SHIFT	0
+#define PAGE_CACHEABLE_SHIFT		1
+#define PAGE_NOT_PRESENT_SHIFT		2
```
jermar: So there is cacheable and not_cacheable shifts, but only not_present shift.

kernel/generic/include/mm/mm.h
```diff
@@ -35,31 +35,116 @@
...
+ * For non-leaf entries, PAGE_USER means that all of the lower-level pages
+ * are PAGE_USER, likewise with PAGE_KERNEL. Exactly one of these two must be
+ * used for leaf entries, but it may be omitted for non-leaf entries.
+ */
+#define PAGE_USER                 (1 << PAGE_USER_SHIFT)
+#define PAGE_KERNEL               (1 << PAGE_KERNEL_SHIFT)
+
+
+
+static inline bool PAGE_FLAGS_VALID(unsigned flags) {
```
jermar: Is this used anywhere?

kernel/generic/include/mm/page.h
```diff
@@ -48,7 +48,7 @@ typedef struct {
 	void (*mapping_insert)(as_t *, uintptr_t, uintptr_t, unsigned int);
 	void (*mapping_remove)(as_t *, uintptr_t);
 	bool (*mapping_find)(as_t *, uintptr_t, bool, pte_t *);
-	void (*mapping_update)(as_t *, uintptr_t, bool, pte_t *);
+	void (*mapping_update)(as_t *, uintptr_t, bool, const pte_t *);
```
jermar: Can mapping_find also use const pte_t *?

kernel/generic/include/mm/mm.h
```diff
@@ -35,31 +35,116 @@
...
+		return flags == (flags & (PAGE_NEXT_LEVEL_PT | PAGE_GLOBAL | PAGE_USER | PAGE_KERNEL));
+
+	/* Leaf entries only. */
+
+	/* Check that at least one permission is set. */
+	if (!(flags & (_PAGE_READ | _PAGE_WRITE | _PAGE_EXEC)))
+		return false;
+
+	/* Check that write implies read. */
+	if ((flags & _PAGE_WRITE) && !(flags & _PAGE_READ))
```
jermar: Seems like PAGE_READ_WRITE will be invalid?

kernel/generic/include/mm/mm.h
jermar: Also, can flags be const?

kernel/genarch/src/mm/page_ht.c
```diff
@@ -60,7 +60,7 @@ static void ht_remove_callback(ht_link_t *);
 static void ht_mapping_insert(as_t *, uintptr_t, uintptr_t, unsigned int);
 static void ht_mapping_remove(as_t *, uintptr_t);
 static bool ht_mapping_find(as_t *, uintptr_t, bool, pte_t *);
```
jermar: const pte_t *?

kernel/generic/include/mm/mm.h
jermar: Oops, my bad, please disregard.

le-jzr:
I might revisit this later. Right now I have other priorities.

PR #30, opened 2018-04-15
Replace autogen.py with something simpler.

le-jzr:
Instead of generating headers for certain structures, the headers are written manually and we automate checking correctness instead. Checking is performed by generating a C source with a bunch of static asserts, using a simple awk script. This is then treated as a normal source file.

The primary motivation for this change is to reduce the complexity of the build process. Also, the `.ag` files we used previously are more difficult to understand than regular C code, and at least one IDE (GNOME Builder) completely refuses to open them.


kernel/arch/amd64/include/arch/boot/memmap_struct.h
```diff
@@ -0,0 +1,46 @@
...
+ * IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT
+ * NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF
+ * THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
```
jermar: Really? :-)

le-jzr: Darn, my evil schemes foiled. :) Shall I uglify those? :-P

le-jzr: Done.

tools/autocheck.awk
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/awk -f
+
```
jermar: Missing copyright header.

le-jzr: Oops, forgot.

PR #31, opened 2018-04-30
Revive support for `-helenos` target specification.

le-jzr:
Instead of maintaining patches separately, directly pull sources from git.
The target is, at the moment, still opt-in.

### DIFF
--- a/tools/toolchain.sh
+++ b/tools/toolchain.sh
@@ -339,7 +339,7 @@ build_target() {
 		--with-gnu-as \
 		--with-gnu-ld \
 		--disable-nls \
-		--enable-languages=c,objc,c++,obj-c++,go \
+		--enable-languages=c,c++,go \
 		--enable-lto \
 		--disable-shared \
 		--disable-werror \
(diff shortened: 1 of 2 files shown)


tools/toolchain.sh
```diff
@@ -388,22 +325,48 @@ build_target() {
...
-		--with-gnu-as --with-gnu-ld --disable-nls --disable-threads \
-		--enable-languages=c,objc,c++,obj-c++ \
-		--disable-multilib --disable-libgcj --without-headers \
-		--disable-shared --enable-lto --disable-werror
+		"--prefix=${PREFIX}" \
+		"--program-prefix=${TARGET}-" \
+		--with-gnu-as \
+		--with-gnu-ld \
+		--disable-nls \
+		--enable-languages=c,objc,c++,obj-c++,go \
```
jermar: Not sure there is still some interest in objc and obj-c++

le-jzr: Good point. I removed them.

PR #36, opened 2018-05-15
ar9271: Set cable state according to connectedness

jermar:
This commit might fix Trac ticket #628. I don't have the hw and don't know how to test this in QEMU (the device seems unsupported as of QEMU 2.12).

### DIFF
--- a/uspace/drv/nic/ar9271/ar9271.c
+++ b/uspace/drv/nic/ar9271/ar9271.c
@@ -165,7 +165,17 @@ static errno_t ar9271_get_device_info(ddf_fun_t *dev, nic_device_info_t *info)
  */
 static errno_t ar9271_get_cable_state(ddf_fun_t *fun, nic_cable_state_t *state)
 {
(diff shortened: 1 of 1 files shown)


jxsvoboda:
Cannot verify. The driver won't bind because it's got 0x0cf3 as vendor ID in its .ma file, but the match for the bust contains 0xcf3. After fixing that the driver crashes with assertion failure upon loading. Apparently there are two attempts at calling ddf_dev_data_alloc(). The immediate cause seems to be because the driver is calling usb_device_create_ddf(). I'm not sure whether this is incorrect or obsolete.. In any case I am not sure how usb+nic driver could have ever worked: libnic allocates and uses dev data (which it absolutely should not do since it's a top-half framework). If USB does the same...

jermar:
Okay, thanks for the info. As USB-NICF clash is now tracked in Trac ticket #730, I am going to merge this now as there is nothing to break.

PR #37, opened 2018-05-15
Fix the return value of memcpy_to/from_uspace().

le-jzr:
Other 64b platforms need to have the assembly part reviewed.


kernel/arch/amd64/src/asm.S
```diff
@@ -78,7 +78,7 @@ FUNCTION_END(memcpy_to_uspace)
 
 SYMBOL(memcpy_from_uspace_failover_address)
 SYMBOL(memcpy_to_uspace_failover_address)
-	xorl %eax, %eax         /* return 0, failure */
```
jermar: Shouldn't really be necessary on amd64 as it will zero-extend the result to the full 64-bit register

le-jzr: Can't hurt either, though.

kernel/generic/src/syscall/copy.c
```diff
@@ -140,12 +140,12 @@ errno_t copy_to_uspace(void *uspace_dst, const void *src, size_t size)
 	ipl = interrupts_disable();
 	THREAD->in_copy_to_uspace = true;
 
-	rc = memcpy_to_uspace(uspace_dst, src, size);
+	rc = !memcpy_to_uspace(uspace_dst, src, size) ? EPERM : 0;
```
jermar: Can this be EOK instead of 0?

jermar:
The other 64-bit platforms look good too. So in fact, no need to change assembly for this PR.

PR #38, opened 2018-05-15
Don't use custom ldscripts in uspace.

le-jzr:
With the memcpy issue identified and fixed, there is no longer a need for IA64 toolchain to be patched.


uspace/srv/loader/Makefile
```diff
@@ -28,8 +28,21 @@
...
 
 USPACE_PREFIX = ../..
+ROOT_PATH = $(USPACE_PREFIX)/..
+CONFIG_MAKEFILE = $(ROOT_PATH)/Makefile.config
 
-EXTRA_LDFLAGS = -Wl,-Ttext-segment=0x70000000
+include $(CONFIG_MAKEFILE)
+
+ifeq ($(UARCH),ia64)
+	# IA64 has a peculiar linker script with a fixed data segment address.
```
jermar: Could such a data segment be move using a linker command line option -Tdata=ORG?

le-jzr: It is, technically, possible, but it would be an architecture-specific and very fragile hack that could break with any binutils release.

le-jzr: Actually, maybe not. I'll need to think on it a bit.

le-jzr: No, there is no clean way to achieve that. It could be broken at any time in a variety of ways. Providing a modified linker script is the only sensible solution I can think of.

le-jzr: What do you mean by "out of sync"? In terms of address range collisions, it is extremely unlikely that any future version of binutils would default to addresses used by the loader here.

PR #39, opened 2018-05-23
[Work in progress] Replace the common case IPC read/write with a single function call.

le-jzr:
Adds three new functions, `async_read()`, `async_write()`, and `async_write_read()`, which implement and encapsulate the usual use cases of data transfer methods. Removes lots of boilerplate.
Updating users of the methods is still work in progress.

### DIFF
--- a/tools/ew.py
+++ b/tools/ew.py
@@ -208,6 +208,9 @@ def qemu_run(platform, machine, processor):
 	if (not 'audio' in cfg.keys()) or cfg['audio']:
 		cmdline += qemu_audio_options()
 
+	if (is_override('nographic')):
+		cmdline += ' -nographic'
+
 	if cfg['image'] == 'image.iso':
 		cmdline += ' -boot d -cdrom image.iso'
 	elif cfg['image'] == 'image.boot':
@@ -381,6 +384,8 @@ def run():
 			overrides['noxhci'] = True
 		elif sys.argv[i] == '-notablet':
 			overrides['notablet'] = True
+		elif sys.argv[i] == '-nographic':
+			overrides['nographic'] = True
 		elif sys.argv[i] == '-qemu_path' and i < len(sys.argv) - 1:
 			expect_qemu = True
 		else:
--- a/uspace/lib/c/generic/async/client.c
+++ b/uspace/lib/c/generic/async/client.c
@@ -1348,7 +1348,14 @@ void *async_as_area_create(void *base, size_t size, unsigned int flags,
 void async_call_begin(async_call_t *call, async_sess_t *sess, sysarg_t imethod,
     sysarg_t arg1, sysarg_t arg2, sysarg_t arg3, sysarg_t arg4)
 {
+	assert(call);
 	memset(call, 0, sizeof(*call));
+	list_initialize(&call->fragments);
+
+	if (!sess) {
+		call->rc = ENOENT;
+		return;
+	}
 
 	call->exch = async_exchange_begin(sess);
 	if (!call->exch) {
@@ -1502,11 +1509,13 @@ void async_call_method_with_finalizer(async_call_t *call,
 	assert(data);
 	data->finalizer = finalizer;
 
-	if (!call->exch)
-		call->rc = ENOENT;
-
 	if (call->rc)
 		return;
+
+	if (!call->exch) {
+		call->rc = ENOENT;
+		return;
+	}
 
 	data->msgid = async_send_fast(call->exch, imethod,
 	    arg1, arg2, arg3, arg4, &data->answer);
@@ -1603,6 +1612,12 @@ void async_call_share_out(async_call_t *call, async_call_data_t *data,
 {
 	async_call_method(call, data,
 	    IPC_M_SHARE_OUT, (sysarg_t) src, 0, (sysarg_t) flags, 0);
+}
+
+void async_call_connect_to_me(async_call_t *call, async_call_data_t *data,
+	sysarg_t arg1, sysarg_t arg2, sysarg_t arg3)
+{
+	async_call_method(call, data, IPC_M_CONNECT_TO_ME, arg1, arg2, arg3, 0);
 }
 
 // TODO: connect me to, connect to me, vfs handle, etc.
--- a/uspace/lib/c/generic/async/client.c
+++ b/uspace/lib/c/generic/async/client.c
@@ -1345,5 +1345,267 @@ void *async_as_area_create(void *base, size_t size, unsigned int flags,
 	return as_area_create(base, size, flags, &pager_info);
 }
 
+void async_call_begin(async_call_t *call, async_sess_t *sess, sysarg_t imethod,
+    sysarg_t arg1, sysarg_t arg2, sysarg_t arg3, sysarg_t arg4)
+{
+	memset(call, 0, sizeof(*call));
+
+	call->exch = async_exchange_begin(sess);
+	if (!call->exch) {
+		call->rc = ENOMEM;
+		return;
+	}
+
+	async_call_method(call, &call->initial, imethod,
+	    arg1, arg2, arg3, arg4);
+}
+
+static suseconds_t time_until(const struct timeval *t)
+{
+	struct timeval tv;
+	getuptime(&tv);
+	if (tv_gteq(&tv, t))
+		return 1;
+
+	return tv_sub_diff(t, &tv);
+}
+
+static errno_t async_call_finish_internal(async_call_t *call, const struct timeval *expires)
+{
+	errno_t rc;
+
+	/* Wait for all the fragments. */
+	while (!list_empty(&call->fragments)) {
+		link_t *tmp = list_first(&call->fragments);
+		async_call_data_t *frag =
+		    list_get_instance(tmp, async_call_data_t, link);
+
+		if (expires) {
+			errno_t trc = async_wait_timeout(frag->msgid, &rc,
+			    time_until(expires));
+			if (trc != EOK)
+				return trc;
+		} else {
+			async_wait_for(frag->msgid, &rc);
+		}
+
+		list_remove(tmp);
+
+		if (rc != EOK)
+			return rc;
+
+		if (frag->finalizer) {
+			rc = frag->finalizer(frag);
+			if (rc != EOK)
+				return rc;
+		}
+	}
+
+	return EOK;
+}
+
+errno_t async_call_finish_timeout(async_call_t *call, const struct timeval *expires)
+{
+	assert(call);
+
+	if (call->rc)
+		return call->rc;
+
+	if (call->exch) {
+		async_exchange_end(call->exch);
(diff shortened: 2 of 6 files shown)


jermar:
Ok, interested where this goes, but my first jerk-knee reaction is: "Wow, functions that take more than 10, sometimes 14 arguments!"

le-jzr:
At first I tried a variadic function, but then I was like "ugh, unchecked argument types".
Most callers don't use most of the arguments, but IMO making a separate function for every combination of required argument counts would be even worse.
I'm also considering just exposing the internal function which has array input. That would mean some boilerplate on the user side, but still much less than the current state.
Overall, it would be best if all the IPC calls were generated from a DSL description, but I'm not sure how much effort that would take.

vhotspur:
I also do not like the idea of function with more than 7 arguments. However the interface for the `async_rw_internal` is not much better IMHO. How about making the `mini-protocol' more explicit and allow named arguments (sort of). Something along the following lines...

```c
typedef enum {
	TYPE_METHOD,
	TYPE_BUFFER_READ,
	TYPE_BUFFER_WRITE,
} mini_ipc_protocol_message_type_t;

typedef struct {
	mini_ipc_protocol_message_type_t type;
	union {
		struct {
			sysarg_t method;
			sysarg_t arg1;
			sysarg_t arg2;
			sysarg_t arg3;
			sysarg_t arg4;
		};
		struct {
			void *buffer;
			size_t size;
			size_t *actual_size;
		};
	};
} mini_ipc_protocol_t;

...
/*
errno_t rc = async_write_2(inetcfg_sess,
	INETCFG_SROUTE_CREATE, 42, 0, 0, 0, &answer,
	dest, sizeof(*dest), NULL,
	router, sizeof(*router), NULL);
*/

mini_ipc_protocol_t protocol[] = {
	{
		.type = TYPE_METHOD,
		.method = INETCFG_SROUTE_CREATE,
		.arg1 = 42,
	},
	{
		.type = TYPE_BUFFER_WRITE,
		.buffer = dest,
		.size = sizeof(*dest),
		.actual_size = NULL
	},
	{
		.type = TYPE_BUFFER_WRITE,
		.buffer = router,
		.size = sizeof(*router),
		.actual_size = NULL
	},
};

errno_t rc = async_protocol(inetcfg_sess, protocol, 3, &answer);
```

Just my $0.02.

jxsvoboda:
My first impressions were:
  - functions with that may arguments - no way
  - why is it just on client side? that makes it asymmetric
  - why embed exchange begin/end? that makes it non-composable

as far as the mini-protocol suggested by vhotspur, I have another idea. Why not construct the protocol steps using procedure calls (and here we are somewhat where we started) kinda like OpenGL display list construction. So the difference from where we are now would be that you would not have to handle the errors, you would just describe the steps and then perform everything in one step.

foo_begin_1(&bar, INETCFG_SROUTE_CREATE, 42);
foo_write(&bar, dest, sizeof(*dest));
foo_write(&bar, router, sizeof(*router));
// foo_end(&bar);

errno_t rc = async_foo(inetcfg_sess, &bar, &answer);

jermar:
When discussing the async API, I would like to mention my crazy idea to try to approximate the futures pattern as much as possible (even though I am not sure how I would do that in C):

<pseudocode that might not make too much sense>
future_t f1 = async_call(...);
future_t f2 = async_call(...);
future_t f3 = async_call(...);
res = async_wait(f1.then(f2).then(f3));

async_wait_all(f1, f2, f3);
f = async_wait_any(f1, f2, f3);

future_t vfs_open();
future_t vfs_read();

async_wait(vfs_open("/tmp/foo").then(vfs_read(...)));
</pseudocode that might not make too much sense>

If we had some well-thought, futures-like primitives like future_t and "then" (or in C rather f3 = async_then(f1, f2)), maybe we could compose the async calls more naturally and also get rid of some of the boilerplate code. Of course, this would be better done in another language than C, but still wanted to mention this.

le-jzr:
I sort of like all of the above proposals, but I'll focus on @jxsvoboda's reply.

> why embed exchange begin/end? that makes it non-composable

That is true. The reason for this is that if the client starts waiting before `async_exchange_end()` is called, then in some cases accidental protocol mismatch might cause both server and client to block indefinitely, instead of returning an error instantly. This I want to avoid, and I agree that the proposed solution is not perfect.

> So the difference from where we are now would be that you would not have to handle the errors, you would just describe the steps and then perform everything in one step.

I like this in concept, but if at all possible, I'd like to avoid making dynamic allocations necessary. Despite current API using dynamic allocations, almost none of the code actually needs it. In fact, the reason I have for bringing this change up is to prepare ground for removing those allocations wherever possible. In this, @vhotspur's suggestion fits better.

I propose a modification of @jxsvoboda's scheme, with provisions for stack memory allocation.
The code might look something like this:

```c
size_t written;
// Stack variables provide storage for bookkeeping information.
async_foo_t bar;
async_foo_fragment_t write1, write2;

// Session is provided in the first call, so that we have more freedom in implementation.
async_foo_begin_1(&bar, inetcfg_sess, INETCFG_SROUTE_CREATE, 42);
async_foo_write(&write1, &bar, dest, sizeof(*dest), NULL);
async_foo_write(&write2; &bar, data, DATA_BUFFER_LEN, &written);
errno_t rc = async_foo_end(&bar, &answer);
```

vhotspur:
> I like this in concept, but if at all possible, I'd like to avoid making dynamic allocations necessary. Despite current API using dynamic allocations, almost none of the code actually needs it. In fact, the reason I have for bringing this change up is to prepare ground for removing those allocations wherever possible.

:+1: 

BTW, my reason for the `mini_ipc_protocol_t[]` was (1) no library-side allocations and (2) making the API more atomic. When the call is built using functions like @jxsvoboda suggested, what came to my mind was, can I insert `if (...) return` between them? Do I need to call some destroy? While if you build the protocol as an array, there is no place to break it as there is only single call. And, obviously, if you do not like the verbosity of the structures, one can always use macros to make it look like this:

```c
mini_ipc_protocol_t protocol[] = {
	MINI_IPC_METHOD_1(INETCFG_SROUTE_CREATE, 42),
	MINI_IPC_WRITE(dest, sizeof(*dest), NULL),
	MINI_IPC_WRITE(router, sizeof(*router), NULL),
	MINI_IPC_END,
};
```

le-jzr:
The array approach doesn't seem very composable to me.
If we go for a flexible generic interface like that, I'd prefer if it worked for all protocols we have, not just the simple cases. Possibly the most complicated in current code, but still quite reasonable, is the case where the caller receives a list of strings of unknown lengths. For that, you first need to wait for one part of the exchange to finish, so that the caller can allocate memory for the strings. With separate functions, you can include a fence function to the elementary functionality, and implement any kind of protocol in a modular way.

le-jzr:
Here's a draft of what I think is a decent API.
It is meant as a total replacement for the current `async_exch_t`-based API, and serves the same function, the primary differences being:
  - lazy error handling (and correspondingly less error handling code on user's side)
  - instead of waiting for individual answers, you wait for every in-flight message to be answered.
  - no support for "forgetting" (i.e. leaking) message information, but "abort" functionality is assumed, which will need kernel support if we want timeouts to work properly.
  - no need for dynamic allocation, the API is ready to put everything on the stack
  - less synchronization is needed for this API than for the current API (thanks to forget being replaced by abort)

le-jzr:
@vhotspur @jermar @jxsvoboda any comments?

vhotspur:
I like it. One suggestion, though: how about that `_begin` would only have two arguments (session and call) and one would need to call `_method` explicitly. It would remove the need for the `.initial` field which is kind of an anomaly wrt. to subsequent calls.

It would be nice if before merge there would be proper Doxygen comments (at least `@brief`s) and (especially) update of http://www.helenos.org/wiki/AsyncSessions.

le-jzr:
The initial method is kind of anomalous in more than that. There's also the `async_call_wait()` call which explicitly waits for everything except that initial method, which is how existing protocols behave. Internally, there's the unavoidable difference that the first message opens the exchange, whereas the following messages are routed as part of that existing exchange. The way I see it, they are just not the same thing, so it would be confusing to make them the same in the API, IMO.

I'll see what I can do about the documentation.

jermar:
In the context of the current async_forget, forgetting means ignoring the reply when it arrives. The arrival is guaranteed eventually (so in this regard the userspace call resources are freed at the same time regardless of whether it is waited for or forgotten). The kernel internally also forgets calls, basically by donating them to the callee and not waiting for them when it terminates. There is no way to invoke this kernel forget from uspace. One possibility would be to have a capability for all sent calls.

le-jzr:
@jermar What do you mean by "arrival is guaranteed eventually"? How can a reply arrive if the server doesn't send it?

jermar:
All non-forgotten calls are eventually answered, at the latest when the task withholding the answer is killed. But if it withholds the call, the proposed abort would wait for it indefinitely.

le-jzr:
The proposed abort is that you tell kernel that you don't want answer to the message any more, and the kernel tells the server. That there is no such operation in the current implementation is nothing new, or even particularly relevant.

le-jzr:
Besides, with an extra indirection, you don't actually need that to implement the API properly. It's just that the lack of abort/cancel functionality in a system that's built on asynchronous IPC seems like a rather significant oversight.

jermar:
The question is whether it is currently better to wait or forget. The former can block the client forever or until the server is killed. If you forget, the client might run out of memory because the server does not answer, but will not deadlock. The second part was about what needs to be done in order for the kernel to support forget.

le-jzr:
Ah, yes, that is a good question. I have two arguments for waiting: One is that forgetting a read/write operation is unsafe. The buffer itself would need to be part of `amsg_t` to make it safe, which would affect the API, making it more complicated and dependent on dynamic allocations. So I don't see it feasible to even design an API that enables forget to exist. The second is that forget is only useful in combination with timeout, and I'm not aware of any non-test code that currently uses timeouts.

le-jzr:
Well, when it comes to the initial implementation, I _could_ make the `abort` forget what can be safely forgotten and only wait for the reads. It would prevent some optimizations, and be generally inconvenient, but otherwise it would work.

le-jzr:
Ok, so it's sort of difficult to discuss the API without a proper documentation. This should make things simpler: https://gist.github.com/le-jzr/2f79d3edc793e8fb57cb9f1578e54561

jermar:
Going back to the idea of async and futures-like API, if one wanted to implement e.g. async vfs_read(), can the callee just return an async_call_t and the caller simply wait for it when it feels like it?

le-jzr:
Not in the sense of having `async_call_t` as the C return type. You could either pass pointer to the state data from outside (like the proposed functions), or return a `malloc()`'d structure.

jermar:
Also, the name async_call_t might be too close to ipc_call_t, confusing readers to believe there is a 1:1 mapping.

PR #40, opened 2018-06-05
Add notification queue to the async framework.

le-jzr:
Instead of running notification handlers in the same fibril that received it,
forcing us to allocate a new fibril when the handler blocks, we instead queue
the notifications, and allow an arbitrary but fixed number of dedicated fibrils
handle them.

Although a service can increase the number of handler fibrils to reduce latency,
there are now no dynamic allocations due to received notifications.
When the same notification is received again while the older instance is still
in queue, the new notification overwrites the old and increments a counter
of received notifications.

The counter is currently unused, because passing it to the handler would
require extensive changes to user code, but it should be straightforward
to make use of it should the need arise.

### DIFF
--- a/uspace/lib/c/generic/async/server.c
+++ b/uspace/lib/c/generic/async/server.c
@@ -180,7 +180,11 @@ typedef struct {
 
 /* Notification data */
 typedef struct {
-	ht_link_t link;
+	/** notification_hash_table link */
+	ht_link_t htlink;
+
+	/** notification_queue link */
+	link_t qlink;
 
 	/** Notification method */
 	sysarg_t imethod;
@@ -188,8 +192,22 @@ typedef struct {
 	/** Notification handler */
 	async_notification_handler_t handler;
 
-	/** Notification data */
-	void *data;
+	/** Notification handler argument */
+	void *arg;
+
+	/** Data of the most recent notification. */
+	ipc_call_t calldata;
+
+	/**
+	 * How many notifications with this `imethod` arrived since it was last
+	 * handled. If `count` > 1, `calldata` only holds the data for the most
+	 * recent such notification, all the older data being lost.
+	 *
+	 * `async_spawn_notification_handler()` can be used to increase the
+	 * number of notifications that can be processed simultaneously,
+	 * reducing the likelihood of losing them when the handler blocks.
+	 */
+	long count;
 } notification_t;
 
 /** Identifier of the incoming connection handled by the current fibril. */
@@ -223,7 +241,13 @@ void async_set_client_data_destructor(async_client_data_dtor_t dtor)
 
 static hash_table_t client_hash_table;
 static hash_table_t conn_hash_table;
+
+// TODO: lockfree notification_queue?
+static futex_t notification_futex = FUTEX_INITIALIZER;
 static hash_table_t notification_hash_table;
+static LIST_INITIALIZE(notification_queue);
+static FIBRIL_SEMAPHORE_INITIALIZE(notification_semaphore, 0);
+
 static LIST_INITIALIZE(timeout_list);
 
 static sysarg_t notification_avail = 0;
@@ -555,7 +579,7 @@ static size_t notification_key_hash(void *key)
 static size_t notification_hash(const ht_link_t *item)
 {
 	notification_t *notification =
-	    hash_table_get_inst(item, notification_t, link);
+	    hash_table_get_inst(item, notification_t, htlink);
 	return notification_key_hash(&notification->imethod);
 }
 
@@ -563,7 +587,7 @@ static bool notification_key_equal(void *key, const ht_link_t *item)
 {
 	sysarg_t id = *(sysarg_t *) key;
 	notification_t *notification =
-	    hash_table_get_inst(item, notification_t, link);
+	    hash_table_get_inst(item, notification_t, htlink);
 	return id == notification->imethod;
 }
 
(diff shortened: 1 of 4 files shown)


le-jzr:
Forgot to mention an important thing!

I added a new synchronization thingy. The issue was that I needed the consumer fibril to block waiting for new notifications in queue. Normally, that would be a job for regular condvar, but a regular condvar would need a mutex lock on the producer side, otherwise the consumer might lose the signal to a race and block on nonempty queue. However, the producer _can't_ lock a fibril mutex because that would cause the very same problem this PR is solving! So I added `fibril_protocvar_t` that only interacts with `async_futex` instead of an arbitrary fibril mutex. I can't rule out there is a simpler solution, so if anyone thinks of anything, let me know.

le-jzr:
Also, the code is currently underdocumented. I'll add documentation if the PR is okayed for merging.

le-jzr:
> So I added `fibril_protocvar_t` that only interacts with `async_futex` instead of an arbitrary fibril mutex.

It occured to me that the same effect (locklessness on the producer side and immunity to lost signals) can be achieved with a completely ordinary counting semaphore.

uspace/lib/c/include/fibril_synch.h
```diff
@@ -90,10 +90,11 @@ typedef struct {
 	fibril_rwlock_t name = FIBRIL_RWLOCK_INITIALIZER(name)
 
 typedef struct {
+	long counter;
 	list_t waiters;
-} fibril_protocvar_t;
+} fibril_semaphore_t;
 
-#define FIBRIL_PROTOCVAR_INITIALIZER(name) \
+#define FIBRIL_SEMAPHORE_INITIALIZER(name) \
```
Aearsis: Shouldn't we initialize the counter here as well?

le-jzr: Good idea!

le-jzr:
Squashed and reordered commits!
Unless I broke something in the reorganization, the semaphore commit is ready to be pushed to mainline independently.

uspace/lib/c/include/fibril_synch.h
```diff
@@ -142,6 +142,29 @@ typedef struct {
 	void *arg;
 } fibril_timer_t;
 
+/**
+ * A counting semaphore for fibrils.
+ * https://en.wikipedia.org/wiki/Semaphore_(programming)
```
jermar: I would drop the wikipedia reference. Everyone knows what a semaphore is.

uspace/lib/c/generic/async/server.c
```diff
@@ -190,6 +194,15 @@ typedef struct {
...
 	/** Notification data */
 	void *data;
+
+	/**
+	 * Data of the most recent notification.
+	 * If count > 1, the data for previous notifications are lost.
+	 */
+	ipc_call_t calldata;
+
+	/** How many of this notification arrived since it was last handled. */
```
jermar: these notifications?

uspace/lib/c/generic/async/server.c
```diff
@@ -190,6 +194,15 @@ typedef struct {
...
+	long count;
```
jermar: Should this be rather unsigned?

uspace/lib/c/include/fibril_synch.h
le-jzr: Okay.

uspace/lib/c/generic/async/server.c
le-jzr: I feel like that wouldn't make it clear enough that I'm talking about multiple instances of the same notification (imethod). I'll try rephrasing it.

uspace/lib/c/generic/async/server.c
le-jzr: I have a rule not to use unsigned unless either the value is a bitfield, or I want wrapping overflows. With signed types, you can make unexpected overflows fatal using `-ftrapv`, or detect them with undefined behavior sanitizer. No such escape hatch exists for unsigned, since for them overflows are legal.

uspace/lib/c/generic/async/server.c
```diff
@@ -662,33 +686,144 @@ static bool route_call(cap_call_handle_t chandle, ipc_call_t *call)
...
+	notification->handler = handler;
+	notification->data = data;
+
+	fid_t fib = 0;
+
+	futex_lock(&notification_futex);
+
+	if (notification_avail == 0) {
+		/* Attempt to create the first handler fibril. */
+		futex_unlock(&notification_futex);
```
jermar: Why do we unlock the notification futex here?

uspace/lib/c/generic/fibril_synch.c
```diff
@@ -628,5 +628,72 @@ fibril_timer_state_t fibril_timer_clear_locked(fibril_timer_t *timer)
 	return old_state;
 }
 
+/**
```
jermar: Can you turn this into a proper doxygen comment?

uspace/lib/c/generic/async/server.c
le-jzr: I'm not entirely sure. Probably just my deep seated discomfort with running expensive operations in a critical section. But it shouldn't make any difference in this case. I'll remove the unlock.

le-jzr:
@jermar thanks for reviewing this! I'll improve the documentation.

PR #42, opened 2018-06-28
[Work in progress] Split fibrils from async framework

le-jzr:
Ok, so this is kind of wonky in some places, but it seems to work fine and I think the primary goal of isolating the two parts from each other has been achieved.
I have plans for further work, but I'm getting pretty fed up with this for now, and don't want to let it rot.

Anyone having strong objections against merging this?

### DIFF
--- a/uspace/lib/c/generic/fibril.c
+++ b/uspace/lib/c/generic/fibril.c
@@ -53,6 +53,7 @@
 #include "private/libc.h"
 
 #define DPRINTF(...) ((void)0)
+#undef READY_DEBUG
 
 /** Member of timeout_list. */
 typedef struct {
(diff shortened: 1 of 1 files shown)


le-jzr:
For reference, some of the wonky parts:
 - `fibril_run_heavy()` could probably use a better name.
 - Currently always creates a thread for async manager. Since loader is a server, the loader's manager thread must be terminated before it exec's the task proper. It's an ugly hack, but some optimization/laziness in the use of threads would fix that.
 - Still no `sys_thread_join()` or `fibril_join()`. Both are needed, but they can wait for a little longer, I think.
 - Still uses `ipc_wait_cycle()`. If we consider the fibril library as an extra layer that replaces threading API by wrapping it, `ipc_wait_cycle()` should, as a syscall that blocks thread, have a fibril-oriented wrapper as well. I.e. there should be something like `fibril_ipc_wait_cycle()` that does the same thing but blocks fibril, not thread. Not an exceptionally difficult thing, but it would introduce some more complexity in the fibril framework. On the plus side, fixing this would probably remove the need for an async manager thread.

le-jzr:
Actually, it's currently horribly slow. I need to fix that.

PR #43, opened 2018-08-24
Get rid of sys/time.h

jermar:
This commit moves the POSIX-like time functionality from libc's
sys/time.h to libposix and introduces C99-like or HelenOS-specific
interfaces to libc.

Specifically, use of sys/time.h, struct timeval, suseconds_t and
gettimeofday is replaced by time.h (C99), struct timespec (C99),
usec_t (HelenOS) and getuptime / getrealtime (HelenOS).


jermar:
There are still some issues with this, so mostly for preview purposes and travis builds.

le-jzr:
As long as we are making this change, we should also fix the weakness of the previous interface, which you've inherited here. It is conflating time _instant_ with time _duration_/_difference_, interpreting the same data type as one or the other depending on context. The functions `ts_add` vs. `ts_add_diff`, or `ts_sub` vs. `ts_sub_diff` exemplify this problem.

The C99 standard library uses `struct timespec` as _instant_ only, and I think we should follow that practice, since we already have `sec_t`/`usec_t`/`nsec_t` as _duration_ types. The relationship should be largely identical to the interaction between `void *` and `ptrdiff_t` -- it makes no sense to add two pointers, but it makes sense to add an offset to a pointer to produce another pointer. Similarly, a difference between two pointers is not a pointer, but an offset.

It would be nice to extend that concept even further and simply not specify any particular "epoch" as the origin of the `struct timespec` value. I.e. in order to determine the _duration_ elapsed since e.g. the UNIX epoch (or any other arbitrary time "landmark"... timemark?), one would have to call a function for that conversion instead of trying to interpret the value itself. Which works really well with the `getuptime` facility, since this way we can consistently track current real time without ever worrying about changes to RTC clock via manual or automatic interventions. All time _instants_ in the system would simply be relative to the boot time, and services like the hardware clock or NTP or whatever would not return the current time, but rather the current best approximation of the calendar time of the epoch.

le-jzr:
Additionally, it would be beneficial to start using "deadlines" instead of "timeouts" in our APIs. Specifying deadline as a time instant when the operation should be interrupted is far superior to specifying "how many (u)seconds" it should run. Particularly if you are implementing a deadline for an operation that is composed from multiple separate operations. Deadline can be passed unchanged to all the suboperations, and the limit will always stay fixed. With timeout, you have to measure how long the operation so far took to subtract it from your desired timeout (or leave it unchanged but allowing the time limit to stretch arbitrarily).

jxsvoboda:
Looks like I didn't notice jzr comment. I completely agree the API needs basically total rework. I am OK with this as the first step, no need to fix the entire world at once....

Yes, confusing time with time difference is the major problem. When you want to have leap seconds, time cannot be reasonably represented as seconds passed since an epoch...

Time should be an opaque type. Adding time with time makes no sense. Functions returning the value of a clock ("current time") should take a clock object as an argument for the sake of testability.

Deadlines sound interesting, it does seem to complicate the code a bit, though..

Btw, note the C11 standard says:
         struct timespec

which holds an interval specified in seconds and nanoseconds (which may represent a calendar time based on a particular epoch);

le-jzr:
> Btw, note the C11 standard says:
struct timespec
>
>which holds an interval specified in seconds and nanoseconds (which may represent a calendar time based on a particular epoch);

Yet in spite of that description, the API pretty much only uses it to represent opaque time. It's actually pretty confusing. Maybe we should just ignore `struct timespec` and use a completely custom time API.

> Deadlines sound interesting, it does seem to complicate the code a bit, though..

I imagine a function like `time_after(duration)` returning `now + duration` in a type suitable for the deadline would alleviate much of that complexity.

jermar:
Merged in a squashed commit bd41ac52cf7d20e9e568c519bf2cb5ac7104b86a

PR #46, opened 2018-10-17
compositor: ignore "mouse button pressed" events when resizing/moving…

maurizio-lombardi:
… a window

If you are moving a window on top of another and you click the
mouse's right button, the compositor brings the background window on top.
This interferes with the window moving operation, leaves
a weird empty rectangle on the screen and confuses the user.

This patch fixes the bug by forcing the compositor to ignore
"mouse button pressed" events until the pointer is released.


le-jzr:
LGTM, thanks! :)

PR #47, opened 2018-10-18
compositor: fix the over operator

maurizio-lombardi:
If a vterm is on top of a vcalc and you decrease the opacity of the
first (or rotate/resize it), the result composite image is obviously wrong
(the vcalc's buttons become black...).

This is because of some mistakes in the over operator's implementation.
This patch fixes the problem by reimplementing it.

### DIFF
--- a/uspace/lib/softrend/compose.c
+++ b/uspace/lib/softrend/compose.c
@@ -52,21 +52,18 @@ pixel_t compose_dst(pixel_t fg, pixel_t bg)
 
 pixel_t compose_over(pixel_t fg, pixel_t bg)
 {
-	uint16_t mf;
-	uint16_t mb;
-
 	uint8_t res_a;
 	uint8_t res_r;
 	uint8_t res_g;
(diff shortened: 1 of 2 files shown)


uspace/lib/softrend/compose.c
```diff
@@ -52,21 +52,18 @@ pixel_t compose_dst(pixel_t fg, pixel_t bg)
...
 pixel_t compose_over(pixel_t fg, pixel_t bg)
 {
-	uint16_t mf;
-	uint16_t mb;
-
 	uint8_t res_a;
 	uint8_t res_r;
 	uint8_t res_g;
 	uint8_t res_b;
+	int res_a_inv;
```
jermar: Can res_a_inv be uint16_t or uint32_t instead of int?

maurizio-lombardi:
Yes it can. Fixed

jermar:
Merged, thanks!

PR #48, opened 2018-10-28
Use user-defined labels instead of phone hashes

jermar:
This commit changes the way how the async framework maps incomming calls
to connections. Instead of abusing the kernel addresses of attached
phones as identifiers, the IPC_M_CONNECT_TO_ME and IPC_M_CONNECT_ME_TO
messages allow the server to specify an arbitrary label which is
remembered in the connected phone and consequently imprinted on each
call which is routed through this phone.

The async framework uses the address of the connection structure as the
label. This removes the need for a connection hash table because each
incoming call already remembers the connection in its label.

To disambiguate this new label and the other user-defined label used for
answers, the call structure now has the request_label member for the
former and answer_label member for the latter.

This commit also moves the kernel definition of ipc_data_t to abi/ and
removes the uspace redefinition thereof. Finally, when forwarding the
IPC_M_CONNECT_TO_ME call, the phone capability and the kernel object
allocated in request_process are now correctly disposed of.


le-jzr:
Nice! I'm sorta tired right now, but I want to review this before it is merged.

jermar:
Thanks. One way to solve that would be to let the async_answer_x() notice that we are answering IPC_M_CONNECT_ME_TO and mix the respective label in automatically.

jermar:
@le-jzr Like this?

le-jzr:
No, sorry, but that's worse.
What I meant is, is there any point in passing the initial `ipc_call_t` into the callback at all? It should not be there. The `async_answer` line should not be there.

jermar:
It has to happen inside the connection handler because we want to give the handler an opportunity to to reject the connection and also other arguments can be returned in a protocol-specific manner. I force-pushed a new version which adds a separate async_accept_0() function.

jermar:
Thanks!

PR #49, opened 2018-10-28
Answer the IPC_M_PHONE_HUNGUP call

jermar:


jermar:
I am going to fix the VFS server loop instead.

PR #50, opened 2018-10-31
Rename THE/the_t to CURRENT/current_t

le-jzr:
Because the word "THE" occurs several times in every licence
header, searching for occurences of "THE" macro is more difficult
than necessary.

While I appreciate the wit of it, using a nonconflicting word
for it is more practical.

### DIFF
--- a/kernel/generic/include/arch.h
+++ b/kernel/generic/include/arch.h
@@ -39,12 +39,11 @@
 #include <config.h>
 
 /*
- * THE is not an abbreviation, but the English definite article written in
- * capital letters. It means the current pointer to something, e.g. thread,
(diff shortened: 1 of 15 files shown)


kernel/generic/include/arch.h
```diff
@@ -39,20 +39,19 @@
 #include <config.h>
 
 /*
- * THE is not an abbreviation, but the English definite article written in
- * capital letters. It means the current pointer to something, e.g. thread,
- * processor or address space. Kind reader of this comment shall appreciate
- * the wit of constructs like THE->thread and similar.
+ * The context pointer. The current_t structure holds pointers to various parts
```
jermar: I would delete the first sentence.

PR #51, opened 2018-11-06
Replace as_area_btree with ordered dictionary.

jxsvoboda:
Anybody else interested &&/|| willing to review this change just let me know.

### DIFF
--- a/kernel/generic/src/mm/as.c
+++ b/kernel/generic/src/mm/as.c
@@ -323,23 +323,30 @@ as_area_t *as_area_next(as_area_t *cur)
  * @param area    Area against which we are testing.
  *
  * @return True if the two areas conflict, false otherwise.
- *
  */
 NO_TRACE static bool area_is_conflicting(uintptr_t addr,
     size_t count, bool guarded, as_area_t *area)
 {
 	assert((addr % PAGE_SIZE) == 0);
 
-	/* Add guard page size unless area is at the end of VA domain */
 	size_t gsize = P2SZ(count);
-	if (guarded && !overflows(addr, P2SZ(count)))
-		gsize += PAGE_SIZE;
+    	size_t agsize = P2SZ(area->pages);
 
-	/* Add guard page size unless area is at the end of VA domain */
-	size_t agsize = P2SZ(area->pages);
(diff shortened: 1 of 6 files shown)


kernel/generic/src/mm/as.c
```diff
@@ -313,76 +377,39 @@ NO_TRACE static bool check_area_conflicts(as_t *as, uintptr_t addr,
...
+
+		/* Next area */
+		odlink = odict_next(odlink, &as->as_areas);
 	}
 
-	node = btree_leaf_node_right_neighbour(&as->as_area_btree, leaf);
-	if (node) {
-		area = (as_area_t *) node->value[0];
+	/* Next area, if any, is the first with base > than our base address */
+	if (odlink != NULL) {
```
jermar: How does hit work if there was no area with base <= our base address, but there is an area with base > our base address? It seems to me odlink will be NULL and so the branch for the latter will not get executed.

kernel/generic/src/mm/as.c
```diff
@@ -487,33 +482,30 @@ NO_TRACE static uintptr_t as_get_unmapped_area(as_t *as, uintptr_t bound,
...
-	list_foreach(as->as_area_btree.leaf_list, leaf_link, btree_node_t, node) {
-
-		for (btree_key_t i = 0; i < node->keys; i++) {
-			as_area_t *area = (as_area_t *) node->value[i];
+	as_area_t *area = as_area_first(as);
+	while (area != NULL) {
+		mutex_lock(&area->lock);
 
-			mutex_lock(&area->lock);
+		addr = ALIGN_UP(area->base + P2SZ(area->pages), PAGE_SIZE);
```
jermar: area->base + P2SZ(area->pages) is already aligned.

kernel/generic/src/mm/as.c
jxsvoboda: Nice catch, fixed.

kernel/generic/src/mm/as.c
jxsvoboda: I guess if you trust area->base to be aligned, then why not.

PR #53, opened 2018-11-07
Add LICENSE.md

jermar:


jxsvoboda:
I'm not sure this is a good idea just because GitHub would like you to do it. LICENSE.md is, by necessity, inaccurate, since different files have different licenses.

jermar:
This serves three main objectives:

1. have sites such as GitHub and GitLab detect the license and provide a nice logo/link on the main toolbar which also (when detected properly) provides a short interpretation of the license terms; being able to establish with ease what the prevalent license used by the project is apparently increases project's friendliness towards new people
1. be able to find HelenOS by license-type
1. have a sample version of what we mean when we say in the README that HelenOS is licensed under the BSD license; note that it is really intended as an answer to the question "What is the prevalent license used by this project?" as there is no copyright statement in this file




GitHub unfortunately still could not detect this as BSD-3-Clause license due to https://github.com/benbalter/licensee/issues/344.

vhotspur:
GitHub does not support multiple licenses per repo and thus I would prefer not to mislead the readers that we are using only one license.

I like the way it is (was) stated in the README and I agree that our LICENSE file should be modeled more after musl than trying to conform with GitHub rules.

One thing that could be in LICENSE to make it machine-processable could be the [SPDX identification](https://spdx.org/), i.e. `BSD-3-Clause AND GPL-2.0-or-later` (see also [this question](https://softwareengineering.stackexchange.com/questions/304874/declaring-multiple-licences-in-a-github-project)).

Just my $0.02.

PR #54, opened 2018-11-07
Remove RCU and related features

jermar:
RCU and related code is very complex, unmaintained, with failing tests and mostly unused.


kernel/generic/src/synch/futex.c
```diff
@@ -343,53 +208,44 @@ static futex_t *get_and_cache_futex(uintptr_t phys_addr, uintptr_t uaddr)
...
-	} else {
-		/* Another thread of this task beat us to it. Use that mapping instead.*/
-		free(fut_ptr);
-		futex_release_ref_locked(futex);
+	pte_t t;
+	bool found;
 
-		futex_ptr_t *dup = member_to_inst(dup_link, futex_ptr_t, cht_link);
-		futex = dup->futex;
+	found = page_mapping_find(AS, ALIGN_DOWN(uaddr, PAGE_SIZE), true, &t);
```
le-jzr: I feel like this needlessly exposes implementation details. Perhaps we could add a function like `page_mapping_get_paddr()`?

PR #55, opened 2018-11-09
Add missing libs to image.iso

vhotspur:
This PR adds `crt*.o` files and `libgcc.a` to the generated image when `CONFIG_DEVEL_FILES` is set so we can compile again inside HelenOS ([this bug](http://www.helenos.org/ticket/750)) (another PR for harbours will be opened shortly).

The change is rather big because it has to find `libgcc.a` which is somewhere in `$CROSS_PREFIX`.

@le-jzr does it make sense to do it like this? Thanks!


le-jzr:
Thanks for working on this. It would be preferable to cross-build the HelenOS toolchain for native builds, instead of that archaic version we have in coastline. But until someone has time to work on that, I suppose this is a decent stopgap.

PR #56, opened 2018-11-10
Remove kernel support for futexes

jermar:
This PR removes futexes from the kernel and replaces them with a much leaner mechanism based on wait queue made usable from uspace via capabilities. As stated in a554ae8, the replacement is not meant to be used between multiple address spaces. For that, one would have to separate the futex counter and the wait queue handle used by the futex. The counter would need to be shared with the other address space and the waitq capability would have to be preallocated and passed to the other task (which we cannot do yet).


jermar:
I tested this with multithreading enabled in VFS. The system runs fine, but after it allocates the first waitq (as can be observed in kconsole by typing caches), some things start to get slower, such as printing out the list of caches. I will keep looking into this.

uspace/lib/c/generic/private/futex.h
```diff
@@ -146,7 +169,7 @@ static inline errno_t futex_down_composable(futex_t *futex,
 static inline errno_t futex_up(futex_t *futex)
 {
 	if (atomic_fetch_add_explicit(&futex->val, 1, memory_order_release) < 0)
-		return __SYSCALL1(SYS_FUTEX_WAKEUP, (sysarg_t) futex);
```
jermar: Actually, this is not correct, the wakeup might need to allocate the handle here too. But what to do with a failure to allocate it? Going to switch to the allocate-before-decrement in sleep strategy.

kernel/generic/src/synch/syswaitq.c
```diff
@@ -0,0 +1,189 @@
...
+
+/** Initialize the user waitq subsystem */
+void sys_waitq_init(void)
+{
+	waitq_cache = slab_cache_create("waitq_t", sizeof(waitq_t), 0, NULL,
+	    NULL, 0);
+}
+
+/** Clean-up all waitq capabilities held by the exiting task */
+void sys_waitq_task_cleanup(void)
```
le-jzr: So, why do we need a separate cleanup routine for each kobject type, instead of just one generic cleanup?

jermar: Slightly different semantics of the objects. Sometimes you need to do further actions. This should eventually become normalized, but not now.

uspace/lib/c/generic/thread/fibril.c
```diff
@@ -795,7 +795,8 @@ int fibril_test_spawn_runners(int n)
 
 	if (!multithreaded) {
 		_ready_debug_check();
-		futex_initialize(&ready_semaphore, ready_st_count);
+		if (futex_initialize(&ready_semaphore, ready_st_count) != EOK)
+			abort();
```
le-jzr: Random idea: What if instead of aborting on failure, we fall back to a spinlock+thread_yield()?

jxsvoboda:
LGTM

PR #58, opened 2018-11-11
Modify kernel malloc()

le-jzr:
This new implementation places the allocation size in front of the allocated
object, instead of relying on the slab allocator being able to determine source
slab cache for an object. This should improve scalability and help reduce
complexity of the memory management subsystem (further changes coming).

The drawback is more memory consumed by small malloc() allocations, however that
can be mitigated by switching to an API where the user provides known object
size to deallocation (most users know it either statically or from length they
necessarily remember).

### DIFF
--- a/kernel/generic/src/mm/malloc.c
+++ b/kernel/generic/src/mm/malloc.c
@@ -81,15 +81,22 @@ void malloc_init(void)
 	}
 }
 
+static inline bool _is_pow2(size_t x)
+{
+	return (x & (x - 1)) == 0;
+}
+
 static void _check_sizes(size_t *alignment, size_t *size)
 {
 	assert(size);
 	assert(alignment);
 
-	assert(*size > 0);
+	/* Force size to be nonzero. */
+	if (*size == 0)
+		*size = 1;
 
 	/* Alignment must be a power of 2. */
-	assert(__builtin_popcountl(*alignment) <= 1);
+	assert(_is_pow2(*alignment));
 	assert(*alignment <= PAGE_SIZE);
 
 	if (*alignment < alignof(max_align_t))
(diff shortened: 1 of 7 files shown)


jermar:
I don't know if improvements in this area are worth the effort - ideally majority of kernel allocations should use a dedicated slab cache.

kernel/generic/src/mm/malloc.c
```diff
@@ -0,0 +1,226 @@
...
+
+	for (i = 0, size = (1 << SLAB_MIN_MALLOC_W);
+	    i < (SLAB_MAX_MALLOC_W - SLAB_MIN_MALLOC_W + 1);
+	    i++, size <<= 1) {
+		malloc_caches[i] = slab_cache_create(malloc_names[i], size, 0,
+		    NULL, NULL, SLAB_CACHE_MAGDEFERRED);
+	}
+}
+
+static inline bool _is_pow2(size_t x)
```
jermar: Could you use ispwr2 from kernel/generic/include/macros.h?

vhotspur:
> This should improve scalability

Do you have any data on this or it is only a guess? Because I have seen quite a lot of changes where the authors thought this "must improve" performance but in reality the performance was degraded...


> switching to an API where the user provides known object size to deallocation

Which would be probably easier because after my quick scan through kernel use of `malloc` is seems that all allocations are either permanent, size is kept too (most generic data structures) or they use `sizeof`.

le-jzr:
I never said it has performance benefit. The purpose is to lift a restriction on slab and frame allocator implementation.

vhotspur:
> I never said it has performance benefit.

Aha, I understood _This should improve scalability_ that it talks about performance ;-)

le-jzr:
Ah, yes, it's pretty ambiguous. What I meant is that by avoiding the need to determine parent cache, we can have per-cache lock granularity, instead of contending for global lock in `frame_get_parent()`. Whether or not that makes slab and malloc frees any faster is... uncertain at best. Probably not. But it simplifies things in ways that are not immediately obvious at this stage.

PR #59, opened 2018-11-12
Remove realmode VESA code

le-jzr:
This simply enables framebuffer setup via multiboot1
(multiboot2 already did it), and removes the obsolete code.


jxsvoboda:
How did you test multiboot1 changes? Can you boot multiboot1 with GRUB2?

Not being a GRUB fan I have a long-term plan to provide an alternative way of booting on BIOS and EFI. While I have nothing against supporting multiboot, I don't think installing GRUB along the system is the best option

As such this change goes somewhat against my plans. On the other hand I agree the current situation where the VESA and multiboot setup is written in assembly is messy and probably would have to be rewritten anyway.

le-jzr:
multiboot1 is still the default, actually.
Ideally, even with alternative to GRUB, most of initialization should be done in the bootloader, including framebuffer setup.

kernel/genarch/src/multiboot/multiboot.c
```diff
@@ -167,6 +168,31 @@ void multiboot_info_parse(uint32_t signature, const multiboot_info_t *info)
...
 		multiboot_memmap(info->mmap_length,
 		    (multiboot_memmap_t *) MULTIBOOT_PTR(info->mmap_addr));
+
+#ifdef CONFIG_FB
+
+	/* Initialize framebuffer. */
+	if ((info->flags & MULTIBOOT_INFO_FLAGS_FB) != 0) {
+		if (info->framebuffer_type != 1) {
+			/* Can't use this framebuffer. */
+			// FIXME: framebuffer_type == 2 is EGA mode, we should be able to use that.
```
jermar: Why not do that if also CONFIG_EGA is enabled?

le-jzr: Because nobody did it yet. We don't even need CONFIG_EGA as a configuration option, if the EGA framebuffer is properly configurable at runtime.

jermar: Well we need it at least to control what support is the kernel built with.

PR #62, opened 2018-11-21
Fix memory leaks

maurizio-lombardi:
Fix a few memory leaks in fontviewer and the font library

### DIFF
--- a/uspace/app/fontviewer/fontviewer.c
+++ b/uspace/app/fontviewer/fontviewer.c
@@ -169,25 +169,25 @@ static errno_t draw(void)
 	source_t descender_fg = rgb(85, 212, 0);
 	source_t leading_bg = rgb(170, 238, 255);
 	source_t leading_fg = rgb(0, 170, 212);
+	font_t *info_font = NULL;
+	font_t *font = NULL;
 
-	font_t *font;
 	errno_t rc = create_font(&font, points);
 	if (rc != EOK) {
 		printf("Failed creating font\n");
-		return rc;
+		goto out_err;
 	}
 
-	font_t *info_font;
 	rc = embedded_font_create(&info_font, 16);
 	if (rc != EOK) {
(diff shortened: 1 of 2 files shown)


uspace/app/fontviewer/fontviewer.c
```diff
@@ -238,6 +238,7 @@ static errno_t draw(void)
 	}
 
 	font_release(font);
+	font_release(info_font);
```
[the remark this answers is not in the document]
maurizio-lombardi: Oh yes, right, and there is another one on line 184 too, will fix them.

uspace/lib/draw/font/bitmap_backend.c
```diff
@@ -211,6 +211,7 @@ static void bb_release(void *backend_data)
 	free(data->glyph_cache);
 
 	data->decoder->release(data->decoder_data);
+	free(data);
```
jermar: Not sure if this is a good location for freeing backend_data for symmetry reasons. font_create() is already passed allocated backend_data, so shouldn't the free go after the call to release in font_release()?

maurizio-lombardi: I am not sure, I mean, font_release() does the following:

``font->backend->release(font->backend_data);``

Isn't the backend's release() callback supposed to fully free the backend's data?

jermar: That's the question. I'd say yes if the backend allocated the data, but that does not seem to be the  case. Or at least the ownership should be better documented. Can @martin-sucha help?

maurizio-lombardi: The backend_data is allocated in bitmap_font_create() (bitmap_backend.c line 231), so the ownership belongs to the backend.

I think we should consider adding the a pointer to backend_data directly into the font_backend structure.

PR #63, opened 2018-11-22
Fix devman function reference counting

jermar:
After commit 498ced18a4, create_fun_node() returns a fun pointer with an
implicit reference. Adding an extra reference for creation thus adds a
reference that will never be dropped and the function object will be
leaked.

This commit fixes the reference counting issue and also adds the missing
check to the call to insert_fun_node().


jermar:
Devman also has a similar issue with device references, however, I am still debugging the fix.

PR #64, opened 2018-11-22
Fix endpoint reference counting

jermar:
After commit 498ced1, an endpoint is created with an implicit reference.
Adding an extra reference for creation thus adds a reference that will
never be dropped and the endpoint will be leaked.

This commit removes the extra reference.


jermar:
Could also @Aearsis have a look?

PR #66, opened 2018-11-22
Fix function reference counting

jermar:
After commit 498ced1, a function is created with an implicit reference.
Adding an extra reference for creation thus adds a reference that will
never be dropped and the function will be leaked.

This commit removes the extra reference.


jxsvoboda:
Note that there is missing a call to refcount_init in create_function.

le-jzr:
> Note that there is missing a call to refcount_init in create_function.

I think I initially built in an assumption that zeroed out refcount is always the correct initial value.
I later decided that it's a bad idea, but forgot to go back and fix all instances.
Yet later I found out that the C standard doesn't specify values for atomics initialized that way, so it wouldn't be clever-compiler-developer-proof anyway.

I'll fix it.

PR #68, opened 2018-11-27
scheduler: fix compile error in kcpulb() when KCPULB_VERBOSE is defined.

maurizio-lombardi:
"t" has been renamed to "thread"


maurizio-lombardi:
I am not sure I still have the commit bit, I have to try.

PR #136, opened 2018-11-30
Ls human readable

matthieuriolo:
Implementeds the options -1/--single-column and -w/--well-formatted. The last option would be -h/--human-readable in linux ls but since the -h option is already in use I choose to use -w

### DIFF
--- a/uspace/app/bdsh/cmds/modules/ls/ls.c
+++ b/uspace/app/bdsh/cmds/modules/ls/ls.c
@@ -65,8 +65,8 @@ static struct option const long_options[] = {
 
 /* Prototypes for the ls command, excluding entry points. */
 static unsigned int ls_start(ls_job_t *);
-static bool ls_print(struct dir_elem_t *);
-static bool ls_print_single_column(struct dir_elem_t *);
+static errno_t ls_print(struct dir_elem_t *);
+static errno_t ls_print_single_column(struct dir_elem_t *);
 static int ls_cmp_type_name(const void *, const void *);
 static int ls_cmp_name(const void *, const void *);
 static signed int ls_scan_dir(const char *, DIR *, struct dir_elem_t **);
@@ -95,7 +95,7 @@ static unsigned int ls_start(ls_job_t *ls)
  *
  * @param de		Directory element.
  */
-static bool ls_print(struct dir_elem_t *de)
+static errno_t ls_print(struct dir_elem_t *de)
 {
 	int width = 13;
 
@@ -109,24 +109,25 @@ static bool ls_print(struct dir_elem_t *de)
 			cap_simplify(&cap);
 
 			char *rptr;
-			if (cap_format(&cap, &rptr) == EOK) {
+			errno_t rc = cap_format(&cap, &rptr);
+			if (rc == EOK) {
 				char bytes[9], suffix[3];
 				sscanf(rptr, "%s %s", bytes, suffix);
 				free(rptr);
 
 				printf("%-40s\t%*s %2s\n", de->name, width - 3, bytes, suffix);
 			} else
-				return false;
+				return rc;
 		}
 	} else if (de->s.is_directory)
 		printf("%-40s\t%*s\n", de->name, width, "<dir>");
 	else
 		printf("%-40s\n", de->name);
 
-	return true;
+	return EOK;
 }
 
-static bool ls_print_single_column(struct dir_elem_t *de)
+static errno_t ls_print_single_column(struct dir_elem_t *de)
 {
 	if (de->s.is_file) {
 		printf("%s\n", de->name);
@@ -134,7 +135,7 @@ static bool ls_print_single_column(struct dir_elem_t *de)
 		printf("%s/\n", de->name);
 	}
 
-	return true;
+	return EOK;
 }
 
 /** Compare 2 directory elements.
@@ -252,7 +253,7 @@ static signed int ls_scan_dir(const char *d, DIR *dirp,
 	}
 
 	for (i = 0; i < nbdirs; i++) {
-		if (!ls.printer(&tosort[i])) {
+		if (ls.printer(&tosort[i]) != EOK) {
 			cli_error(CL_ENOMEM, "%s: Out of memory", cmdname);
 			goto out;
 		}
@@ -469,7 +470,7 @@ int cmd_ls(char **argv)
 	scope = ls_scope(de.name, &de);
 	switch (scope) {
 	case LS_FILE:
-		if (!ls.printer(&de)) {
+		if (ls.printer(&de) != EOK) {
 			cli_error(CL_ENOMEM, "%s: Out of memory", cmdname);
 			return CMD_FAILURE;
 		}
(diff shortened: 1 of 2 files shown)


jxsvoboda:
Just FYI, we're not trying to emulate Linux. The fact that the author of Bdsh made it very much a UNIX shell is unfortunate and I consider Bdsh to be slated for eventual replacement by something else. In any case if you are working on it, don't feel obligated to continue copying things from Linux, rather try to do the right thing (T.M.) So no need to use same options as on Linux or even have same behavior. From my point of view the human readable version should be the default and the exact sizes should be an option.

Now down to business:
 - with -w the size column should be right-justified, not left-justified
 - we may need to deal with the problem that 'B' is shorter than 'kB' and does not align properly

The -1 mode behaves strangely: there is no way to see if an entry is a directory or file, yet directories are sorted first. Either dirs should be marked with a trailing slash (dir/) or all entries should be sorted just alphabetically.

jxsvoboda:
OK with the way capacity is formatted I'm actually not sure which (if any) justification is correct although I'd probably prefer right-justified so that all the B/kB/MB align (even if the numbers can never align correctly). I do see it's left-justified for exact sizes as well (and should be right-justified) so it's another problem - but, can we do something about it? You might want to have a look at io/table.h for tabular output, although it does not support right justification at the moment.

uspace/app/bdsh/cmds/modules/ls/ls.c
```diff
@@ -89,14 +97,45 @@ static unsigned int ls_start(ls_job_t *ls)
...
-		printf("%-40s\t%llu\n", de->name, (long long) de->s.size);
-	else if (de->s.is_directory)
-		printf("%-40s\t<dir>\n", de->name);
+	int width = 13;
+
+	if (de->s.is_file) {
+		if (ls.well_formatted) {
+			cap_spec_t cap;
+
+			cap_from_blocks(1, de->s.size, &cap);
```
jxsvoboda: If anything, this should be cap_from_blocks(de->s.size, 1, &cap) - the file size can be 64-bit, while size_t can only hold 32-bits on 32-bit CPU.

uspace/app/bdsh/cmds/modules/ls/ls.h
```diff
@@ -25,4 +18,15 @@ struct dir_elem_t {
 	vfs_stat_t s;
 };
 
+typedef struct {
+	/* Options set at runtime. */
+	unsigned int recursive;
+	unsigned int sort;
+
+	bool single_column;
+	bool well_formatted;
```
jxsvoboda: Please rename to exact_size and invert the logic to match the -e option (i.e. exact_size is true if -e is provided),

uspace/app/bdsh/cmds/modules/ls/ls.c
```diff
@@ -89,14 +97,45 @@ static unsigned int ls_start(ls_job_t *ls)
...
+
+			char *rptr;
+			if (cap_format(&cap, &rptr) == EOK) {
+				char bytes[9], suffix[3];
+				sscanf(rptr, "%s %s", bytes, suffix);
+				free(rptr);
+
+				printf("%-40s\t%*s %2s\n", de->name, width - 3, bytes, suffix);
+				
+				//if there is a failure with cap_format we simply print out an unformatted size
```
jxsvoboda: cap_format will only fail if we run out of memory. In general, OOM errors should be handled as follows: print an error message "Out of memory.", stop what you are doing and return to the closest interactive loop. In this case you should print "Out of memory.", stop listing the directory and return a non-zero exit code from ls.

matthieuriolo:
I'm not sure why it still shows that there are changes requested. But I took care of them

le-jzr:
@jxsvoboda You haven't replied since your comments were addressed. Are you satisfied with this state?

uspace/app/bdsh/cmds/modules/ls/ls.c
```diff
@@ -87,14 +95,46 @@ static unsigned int ls_start(ls_job_t *ls)
  *
  * @param de		Directory element.
  */
-static void ls_print(struct dir_elem_t *de)
+static bool ls_print(struct dir_elem_t *de)
```
le-jzr: Please don't use `bool` return type as an error indicator.
I know some other code uses this convention, but it tends to obscure meaning, so we should be trying to reduce it.
The two "good" options you have is to either use `errno_t` and return `<errno.h>` constants as appropriate, or to use the `CL_E...` constants resident in bdsh.

uspace/app/bdsh/cmds/modules/ls/ls.c
```diff
@@ -87,14 +95,46 @@ static unsigned int ls_start(ls_job_t *ls)
...
+		} else {
+			cap_spec_t cap;
+
+			cap_from_blocks(de->s.size, 1, &cap);
+			cap_simplify(&cap);
+
+			char *rptr;
+			if (cap_format(&cap, &rptr) == EOK) {
+				char bytes[9], suffix[3];
+				sscanf(rptr, "%s %s", bytes, suffix);
```
le-jzr: Using `sscanf()` here is an overkill. It's also unsafe. The static arrays with magic constant length may cause memory corruption, because you didn't inform `sscanf()` how long they are, and the behavior of `cap_format()` can change. Any place with arrays such as these is a secret ticking bomb. Avoid it whenever at all possible. It's best to avoid any variant of `scanf()` as well, if you can help it. They are hard to use correctly.

Instead, you can find the last space with `strrchr()`, overwrite it with `'\0'`, and use the two parts of the buffer as separate strings.

uspace/app/bdsh/cmds/modules/ls/ls.c
```diff
@@ -109,32 +109,33 @@ static bool ls_print(struct dir_elem_t *de)
 			cap_simplify(&cap);
 
 			char *rptr;
-			if (cap_format(&cap, &rptr) == EOK) {
+			errno_t rc = cap_format(&cap, &rptr);
+			if (rc == EOK) {
```
le-jzr: Could you also invert this condition so that the following code isn't unnecessarily indented? Thanks!

uspace/app/bdsh/cmds/modules/ls/ls.c
```diff
@@ -87,14 +95,52 @@ static unsigned int ls_start(ls_job_t *ls)
...
 {
-	if (de->s.is_file)
-		printf("%-40s\t%llu\n", de->name, (long long) de->s.size);
-	else if (de->s.is_directory)
-		printf("%-40s\t<dir>\n", de->name);
+	int width = 13;
+
+	if (de->s.is_file) {
+		if (ls.exact_size) {
+			printf("%-40s\t%*llu\n", de->name, width, (long long) de->s.size);
```
le-jzr: Also, early return here would avoid another level of indentation.

matthieuriolo: I dont understand what you mean. Every case needs to be checked -> there is no way to "return early"

le-jzr: After this `printf()`, there is no other code present until the end of function. Only `else` branches. Therefore, you can put `return EOK;` right after the `printf()`. When you do that, you don't need `else` after it, so you can reduce indentation of the block of code below.

Horizontal space is a premium currency, so any opportunity to reduce the amount of indentation used is a good thing.

matthieuriolo:
> Awesome, thanks!

Very welcome :)

PR #140, opened 2018-12-08
mkext4: fix infinite loop when passing an invalid argument.

maurizio-lombardi:
Also adds the --help parameter.

### DIFF
--- a/uspace/app/mkext4/mkext4.c
+++ b/uspace/app/mkext4/mkext4.c
@@ -123,10 +123,19 @@ int main(int argc, char **argv)
 			++argv;
 		}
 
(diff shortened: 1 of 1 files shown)


maurizio-lombardi:
@jxsvoboda what is this condition used for? I don't understand it.

``
if (str_cmp(*argv, "-") == 0)
``
I didn't remove it because I don't want to break things.

jxsvoboda:
The sole dash is meant to be used as terminator for the option list. But it should have been "--" to be compatible with getopt.

maurizio-lombardi:
Self-NACK, I realized I did a mistake and I have to resubmit the patch

PR #141, opened 2018-12-10
mkext4: fix infinite loop when passing an invalid argument.

maurizio-lombardi:
Also adds the --help parameter.

### DIFF
--- a/uspace/app/mkext4/mkext4.c
+++ b/uspace/app/mkext4/mkext4.c
@@ -67,7 +67,7 @@ int main(int argc, char **argv)
 	--argc;
 	++argv;
 
-	while (*argv[0] == '-') {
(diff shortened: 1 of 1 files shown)


uspace/app/mkext4/mkext4.c
```diff
@@ -86,6 +86,7 @@ int main(int argc, char **argv)
 
 			--argc;
 			++argv;
+			continue;
```
jxsvoboda: If this was the last argument, then *argv is now NULL and while (*argv[0] == '-') will crash. Need to add *argv != NULL *&& to the while condition.

maurizio-lombardi: You're right, fixed.

PR #143, opened 2018-12-13
Ext4 fixes

maurizio-lombardi:
Fixes a few bugs in the ext4 library and allows the user to create ext2 filesystems with larger block sizes


maurizio-lombardi:
@jxsvoboda I tested mkext4 with different device sizes and 1, 2 and 4k blocks. It always passed fsck.ext2 test.

PR #146, opened 2018-12-28
fixes #779

matthieuriolo:


jermar:
Thanks for working on this.
Could you please change the commit message to indicate what it fixes instead of the reference to the ticket number? The ticket numbers are not necessarily stable in time, so having a textual description is preferable.
As for the change itself, I suspect the bug started to manifest with a realloc() fix of commit d8023313b00bbb8e958c404ffdc81f4bf36d7f00.

le-jzr:
That is quite likely.
`realloc()`'s behavior with zero size is implementation-defined, and I've since come to question the wisdom of the change I made. It might be safer to make it possible for `malloc()` and `realloc()` to work with zero-sized allocations. The simplest way would be to just automatically change size 0 to size 1.

matthieuriolo:
@jermar: Yes I can do that. I'm just really used to it. Bad habit

> The simplest way would be to just automatically change size 0 to size 1.

Definitely, since the return value NULL usually indicates an error for alloc functions

jermar:
However, let's not fiddle with realloc() at this point.

jermar:
Sorry for nitpicking, but now the commit message is slightly confusing. I would write something like:

```
Prevent ls -r from crashing inside an empty directory

Our implementation of realloc() behaves like free() and returns NULL if size is 0.
Invoking it with size 0 inside ls_scan_dir() leads to a double free (and displaying
of an error message). This commit therefore avoids invoking realloc() when the
size would be 0 because of zero nbdirs.
```
Btw, here is a [wiki article](http://www.helenos.org/wiki/GitCommitRules) on writing commit messages for HelenOS.

le-jzr:
Sorry for another nitpick, but the use of "current implementation of `realloc()`" makes it unclear why it's an issue of the changed code and not an issue of `realloc()`. I'd add a mention that the standard makes this behavior implementation-defined, so the original code is not correct standard-wise.

jermar:
Returning NULL from realloc() causes the function _only_ to print the error message and bail out early, while behaving as free() leads to the double-free later in ls_recursive(), which is imho more serious and yet is not mentioned in the commit message (the double-free is not because realloc() returns NULL, but because it calls free() beforehand).

jermar:
Thanks again!

PR #147, opened 2018-12-29
Ticket 684

matthieuriolo:
As mention in ticket 684 bdsh would allow to execute cmds directly from the working directory without the use of "./". However, the cmd would not show up in the auto-completion. This request adds an extra test in find_command() to ensure only relative or absolute pathes get passed to try_access() > vfs_lookup_open() > vfs_lookup() > vfs_absolutize()

sidenote: cmd can contain a / which allows them to be located in a sub directory in /app
`mkdir /app/sub`
`cp /app/ping /app/sub/p`
`sub/p`

### DIFF
--- a/uspace/app/bdsh/exec.c
+++ b/uspace/app/bdsh/exec.c
@@ -46,9 +46,6 @@
 #include "exec.h"
 #include "errors.h"
 
-/* FIXME: Just have find_command() return an allocated string */
-static char *found;
-
 static char *find_command(char *);
 static int try_access(const char *);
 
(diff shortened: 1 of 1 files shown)


le-jzr:
Sorry about taking this long to review this.
With this patch, there is now an issue that subdirectories of CWD now autocomplete, but fail to run.

matthieuriolo:
> there is now an issue that subdirectories of CWD now autocomplete, but fail to run

They are supposed to fail. But they are not supposed to appear in autocomplete. After some research I noticed that this problem is actually more complicated.

_There is no && notation in bdsh. I simply used it for making the code smaller_

1) `cd / && test` tabtab will correctly not autocomplete contents from test/
2) `cd /test && test` tabtab will correctly not autocomplete contents from test/
3) `cd / && test/` tabtab will INCORRECTLY autocomplete contents from test/
4) `mkdir app/sub && cp app/date app/sub/date && sub/da` tabtab will incorrectly NOT autocomplete content from the search path. But it will correctly autocomplete the directory sub (`su` tabtab)

Good catch [le-jzr](/le-jzr), it appears that the ticket missed the information that autocompletion does work, when the directory path is present in the command line.

matthieuriolo:
I replaced this branch with #163

PR #148, opened 2019-01-06
Added screenshot

adamnemecek:
I think that the README needs screenshot. Razzle and dazzle.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,5 @@
 # HelenOS
 
-![screenshot](http://www.helenos.org/chrome/site/screenshot.png "Screenshot")
-
 HelenOS is a portable microkernel-based multiserver operating
 system designed and implemented from scratch. It decomposes key
 operating system functionality such as file systems, networking,
@@ -10,6 +8,8 @@ fine-grained user space components that interact with each other
 via message passing. A failure or crash of one component does not
 directly harm others. HelenOS is therefore flexible, modular,
 extensible, fault tolerant and easy to understand.
+
+![screenshot](http://www.helenos.org/raw-attachment/wiki/Screenshots/screenshot.png "Screenshot")
 
 HelenOS aims to be compatible with the C11 and C++14 standards, but does not
 aspire to be a clone of any existing operating system and trades compatibility
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # HelenOS
 
+![screenshot](http://www.helenos.org/chrome/site/screenshot.png "Screenshot")
+
 HelenOS is a portable microkernel-based multiserver operating
 system designed and implemented from scratch. It decomposes key
 operating system functionality such as file systems, networking,


README.md
```diff
@@ -1,5 +1,7 @@
 # HelenOS
 
+![screenshot](http://www.helenos.org/chrome/site/screenshot.png "Screenshot")
```
jermar: Maybe the following URL is slightly better: http://www.helenos.org/raw-attachment/wiki/Screenshots/screenshot.png

jermar:
Hi, thanks for contributing to HelenOS. Let's see if there is some feedback from other developers.

README.md
le-jzr: Isn't it possible to display image from the git repo itself?

adamnemecek: It is. Where do I put the image in the repo? Should I create a resources/assets directory in the repo and put the image there?

matthieuriolo:
I changed my mind and had to revoke my thumpsup. Even though it would be lovely to have a screenshot on github, the README is supposed to be readable with a simple texteditor/textviewer. Adding images would end up:

1. making the file less readable (long lines)
2. obfuscating the purpose (first entry file)
3. creates duplicated code (harder to maintain)

It probably would be better to add to the existing list a new link to the webpage where the screenshots are located:

* [Website](http://helenos.org)
* [Screenshots](http://helenos.org/wiki/Screenshots)
* [Wiki](http://helenos.org/wiki)
...

or to add an extra paragraph as the following:

> Please visit our wiki for [screenshots](http://helenos.org/wiki/Screenshots) or our youtube channel for [videos](https://www.youtube.com/channel/UC4vY8mtJaulyi2NEkTMQtow) to see HelenOS in action.


PS: I'm fully aware that there is no official yt channel for helenos. But would be a nice-to-have-in-the-future
PPS: The only exception to the rule could be the logo itself which should be visible on the readme

le-jzr:
While I'd be against using images in general instead of plain text explanation, I think one representative screenshot would not hurt.

README.md
le-jzr: There's some non-code stuff in `contrib/`, so maybe there? @jermar?

jermar: TBH, I am not a big fan of including screenshots into the repository.

adamnemecek:
@matthieuriolo It's a single image. The single line won't change readability. How does including images create duplicated code? It doesn't obscure any purpose, it's literally a markdown link.

You can add a link to images however the clickthrough rates will be minimal. 

There's a lot of operating systems out there but only few that have a custom GUI. You want to show this off, people would love that.

README.md
le-jzr: Ok, an external link is fine then.

matthieuriolo:
@adamnemecek 

> It's a single image. The single line won't change readability.

It kinda does, since the line is longer than 80 chars. I'm also pretty sure that in future people will request to add more images ;) After all, I already recommended to add the logo to the README

> How does including images create duplicated code?
You have to maintain the wikipage with the screenshots and the readme. Twice the work for the same job

> It doesn't obscure any purpose, it's literally a markdown link

I think it does. If you open README with vi you expect a summary, some information about licensing and build manual. Not an image gallery. Everything which is not visible with a text viewer obscures the purpose of a README

https://en.wikipedia.org/wiki/README

> You can add a link to images however the clickthrough rates will be minimal.

I think you underestimate the people here. Especially since you give yourself a reason to why a README is supposed to be short... as shorter it is, as more likely it would be read

> There's a lot of operating systems out there but only few that have a custom GUI. You want to show this off, people would love that.

Let's be honest here, a GUI does not make a good OS ;) there are other characteristics which are more important

vhotspur:
My $0.02: I think README.md should be as plain-text as possible but *one* image is fine. I would only put it below the first paragraph(s) as I believe the general description should be what a user sees first.

adamnemecek:
Ok, I changed the URL and moved it below the first paragraph. You can see it here

https://github.com/adamnemecek/helenos

jermar:
Looks good!

PR #149, opened 2019-01-09
mips32: clobber memory on sycall

vhotspur:
As the syscall may touch arbitrary memory, we need to prevent the compiler to cache some values in registers.

Actual problem was demonstrated on Kalisto during school assignments (thanks Adam Frey) but HelenOS uses basically the same code, thus the issue applies here too.

Can someone better versed with other architectures check that this does not apply to them as well?

### DIFF
--- a/uspace/lib/c/arch/mips32/src/syscall.c
+++ b/uspace/lib/c/arch/mips32/src/syscall.c
@@ -56,11 +56,17 @@ sysarg_t __syscall(const sysarg_t p1, const sysarg_t p2, const sysarg_t p3,
 	      "r" (__mips_reg_t0),
 	      "r" (__mips_reg_t1),
 	      "r" (__mips_reg_v0)
+	    :
 	      /*
 	       * We are a function call, although C
 	       * does not know it.
 	       */
-	    : "%ra"
+	      "%ra",
+	      /*
+	       * Clobber memory too as some arguments might be
+	       * actually pointers.
+	       */
+	      "memory"
 	);
 
 	return __mips_reg_v0;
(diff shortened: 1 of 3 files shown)


le-jzr:
I think all the registers that are not explicitly preserved by the kernel must be marked clobbered/modified as well.

le-jzr:
Does it? Hmm. In general the kernel side of syscall doesn't need to preserve scratch registers. I think we might be doing it for debugging reasons, but still it would be safer (more future-proof) if the userspace portion works even if we decide not to store them and instead just clear them on return (as an optimization).

le-jzr:
Though there should be no expectation that the syscall follows standard calling convention, so that's not really relevant. Ignore me.

uspace/lib/c/arch/mips32/src/syscall.c
```diff
@@ -60,7 +60,12 @@ sysarg_t __syscall(const sysarg_t p1, const sysarg_t p2, const sysarg_t p3,
 	       * We are a function call, although C
 	       * does not know it.
 	       */
-	    : "%ra"
+	    : "%ra",
```
jermar: To make things consistent, the colon should now (that there are two clobbers) go before the comment.

jermar:
IMHO arm32 and ppc32 need the same fix.

jermar:
An interesting case is what happens on architectures that have the syscall wrapper implemented in a .S file in an extern function. From that I'd assume that calling a function represents an implicit memory clobber. It certainly does represent a 'sequence point' from the point of the language

le-jzr:
Yes, external function that's not declared with `__attribute__((pure))` is always assumed to clobber memory.

jxsvoboda:
Am I right thinking that this can only cause problems with LTO? Because otherwise any incorrect optimizations are contained to the syscall function. When the function is used, without LTO, it is an external function and the compiler will assume it clobbers memory. With LTO the linker will have the information that it is not supposed to clobber memory.

le-jzr:
It doesn't follow. Functions are assumed to clobber memory because they are opaque to the caller. When LTO is involved, the compiler can consider internal details of the function. Inline assembler needs the clobber information explicitly, because gcc can't understand the assembly. In principle, this can affect even assembly generated by gcc in prologue/epilogue of otherwise empty function. In practice it doesn't seem to, but it can.

vhotspur:
> IMHO arm32 and ppc32 need the same fix.

Added.

Any objections for merging this?

PR #150, opened 2019-01-09
Minor changes to default configuration

vhotspur:
Any objections for these changes? They are mostly motivated by better coverage in CI:

 * Smaller image means we can pack more ported software to the image for testing (currently, this means only mips32 on MSIM). Obviously, this is only temporary until we can properly boot from disk (but *that* problem exists for several years at least, so this change should not hurt anything, really).
 * Inclusion of PCUT tests means that we can run them regularly too (patch to CI for that is almost ready).


jermar:
lgtm

jxsvoboda:
LGTM

PR #151, opened 2019-01-09
Benchmarking

vhotspur:
This PR refactors quite a bit the `perf` utility that was recently added to benchmark some internals part of HelenOS userspace.

Summary of changes (the diff is unfortunately quite large):
 * Factor out all the duplicate code in individual benchmarks into a common harness
 * Properly compute average throughput
 * Added an awful function to compute square root until proper implementation is available (for the precision we need it is quite sufficient) [sqrt is needed to compute standard deviation]
 * Added option to dump data to CSV for further processing
 * Added simple benchmark for fibril mutexes
 * Merged benchmarks from `bnchmark` application and removed it
    * These benchmarks are FS-related and IMHO are quite badly designed (they really stress only the block cache and not the FS on top of it; but still they might be useful as IPC benchmark of large data volumes)
 * Renamed `perf` to `hbench` to stress that it is a benchmarking application rather than a generic performance measurement tool (compared to `perf` in GNU/Linux)
 * Extended documentation a little bit
 * Added function for measuring wall-clock time (internally pass-through to `uptime` but available for other applications as *the right API* for this in HelenOS)

Any objections for merging this?


uspace/app/hbench/hbench.h
```diff
@@ -0,0 +1,103 @@
...
+/*
+ * So far, a simple wrapper around system stopwatch.
+ * Eventually, we could collection of hardware counters etc. without
+ * modifying signatures of any existing benchmark.
+ */
+typedef struct {
+	stopwatch_t stopwatch;
+} benchmeter_t;
+
+static inline void benchmeter_init(benchmeter_t *meter)
```
jxsvoboda: There is no good reason for these not to be regular functions.

uspace/app/hbench/ipc/ns_ping.c
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+#include <stdio.h>
+#include <ns.h>
+#include <async.h>
+#include <errno.h>
+#include <str_error.h>
+#include "../hbench.h"
+
+static bool runner(benchmeter_t *meter, uint64_t niter,
```
jxsvoboda: You could prefix the name of the function even if it is static.

uspace/app/hbench/params.c
```diff
@@ -0,0 +1,146 @@
...
+static hash_table_ops_t param_hash_table_ops = {
+	.hash = param_hash,
+	.key_hash = param_key_hash,
+	.key_equal = param_key_equal,
+	.equal = param_equal,
+	.remove_callback = param_remove
+};
+
+/** Table of extra parameters (of param_t). */
+static hash_table_t param_hash_table;
```
jxsvoboda: Global state is bad practice. Please change this into an object.

uspace/app/hbench/fs/dirread.c
```diff
@@ -0,0 +1,81 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include "../hbench.h"
+
+/*
+ * Note that while this benchmark tries to measure speed of direct
+ * read, it rather measures speed of FS cache as it is highly probable
+ * that the corresponding blocks would be cached after first run.
+ */
+static bool runner(benchmeter_t *meter, uint64_t size,
```
jxsvoboda: Docblock?

uspace/app/hbench/fs/fileread.c
```diff
@@ -0,0 +1,107 @@
...
+#include "../hbench.h"
+
+#define BUFFER_SIZE 4096
+
+/*
+ * Note that while this benchmark tries to measure speed of file reading,
+ * it rather measures speed of FS cache as it is highly probable that the
+ * corresponding blocks would be cached after first run.
+ */
+static bool runner(benchmeter_t *meter, uint64_t size,
```
jxsvoboda: Docblock?

uspace/app/hbench/hbench.h
```diff
@@ -0,0 +1,103 @@
...
+ */
+
+#ifndef HBENCH_H_
+#define HBENCH_H_
+
+#include <errno.h>
+#include <stdbool.h>
+#include <perf.h>
+
+/*
```
jxsvoboda: Why is this not a docblock?

uspace/app/hbench/main.c
```diff
@@ -0,0 +1,393 @@
...
+
+	while (fabs(estimate - prev_estimate) > precision) {
+		prev_estimate = estimate;
+		estimate = (prev_estimate + value / prev_estimate) / 2.;
+	}
+
+	return estimate;
+}
+
+/*
```
jxsvoboda: Why is this not a docblock?

uspace/app/hbench/main.c
```diff
@@ -0,0 +1,393 @@
...
+	if (duration_usec > 0) {
+		double nanos = stopwatch_get_nanos(&meter->stopwatch);
+		double thruput = (double) workload_size / (nanos / 1000000000.0l);
+		printf(", %.0f ops/s.\n", thruput);
+	} else {
+		printf(".\n");
+	}
+}
+
+/*
```
jxsvoboda: Why is this not a docblock?

.gitignore
```diff
@@ -106,6 +105,7 @@ uspace/app/fdisk/fdisk
 uspace/app/fontviewer/fontviewer
 uspace/app/getterm/getterm
 uspace/app/gunzip/gunzip
+uspace/app/hbench/hbench
```
jxsvoboda: Does 'h' stand for HelenOS? Please do NOT encode name of the project into names of any binaries, tools, etc. What if the name ever needs to change or the tool is ported to another OS? Can we have e.g. just bench?

uspace/app/hbench/doc/doxygroups.h
```diff
@@ -0,0 +1,43 @@
+/** @addtogroup hbench hbench
+ * @brief HelenOS user space benchmarks
```
jxsvoboda: The word HelenOS is redundant

uspace/lib/c/include/perf.h
```diff
@@ -0,0 +1,105 @@
...
+/** @file
+ * System performance measurement utilities.
+ */
+
+#ifndef LIBC_PERF_H_
+#define LIBC_PERF_H_
+
+#include <time.h>
+
+/** Stopwatch is THE way to measure elapsed time on HelenOS. */
```
jxsvoboda: Please remove the word HelenOS. It is redundant.

uspace/lib/c/include/perf.h
```diff
@@ -0,0 +1,105 @@
...
+/** Get elapsed time in nanoseconds.
+ *
+ * @param stopwatch Stopwatch to read from.
+ * @return Elapsed time in nanoseconds.
+ */
+static inline nsec_t stopwatch_get_nanos(stopwatch_t *stopwatch)
+{
+	return ts_sub_diff(&stopwatch->end, &stopwatch->start);
+}
+
```
le-jzr: `stopwatch_set_elapsed` vs `stopwatch_get_nanos`?
Also, the API suggests you can start and stop the watch repeatedly, and then get the aggregate result, which does not match the actual behavior.

vhotspur: > `stopwatch_set_elapsed` vs `stopwatch_get_nanos`?

Good catch, fixed. Thanks.

> Also, the API suggests you can start and stop the watch repeatedly, and then get the aggregate result, which does not match the actual behavior.

I do not think the API suggests it but I added extra comment.

uspace/app/hbench/doc/doxygroups.h
vhotspur: Fixed.

uspace/app/hbench/main.c
vhotspur: Done.

uspace/app/hbench/main.c
vhotspur: Done.

jxsvoboda:
You haven't addressed two of my comments:
- benchmeter should not be inline
- params should not be stored in a global variable
Is that intentional or an omission? I won't force it on you, it's your call.

vhotspur:
> You haven't addressed two of my comments:
>  ...

I did not have time yet to react to it. Hopefully next week.

le-jzr:
So, trying to use `perf` to test performance, it occurred to me that the whole methodology of it is rather bad. If you do a fixed number of iterations, and then read the time it took, we are exposing ourselves to two problems:
  1) The benchmarking will tend to take excessive amount of time for no good reason,
  2) The precision of the measurement is affected by the granularity of system timer, which is rather low, as it's limited by kernel tick frequency.

Now, in my opinion, we need an implementation that runs the benchmark loop until a particular time occurs, i.e. it counts the number of iterations taken before a particular kernel tick. This naturally results in more complex code, as well as some overhead from timer checking, which needs to be accounted for, but would give vastly better results and usability.

vhotspur:
> So, trying to use `perf` to test performance, it occurred to me that the whole methodology of it is rather bad. If you do a fixed number of iterations, and then read the time it took, we are exposing ourselves to two problems:
> 
> 1. The benchmarking will tend to take excessive amount of time for no good reason,

Benchmarking takes time ;-). You need to run it for longer time to better capture the variability (e.g. standard deviation) of the measurements. Honestly, 10s and 10 restarts are quite low numbers.

Since we run on quite wide variety of platforms with vastly different performance, doing the benchmarking this way is actually the only sensible approach. We do a warm-up to determine proper workload size and we quite long interval to avoid the trap of measuring intervals not much bigger than the granularity of the timer.

> 
> 2. The precision of the measurement is affected by the granularity of system timer, which is rather low, as it's limited by kernel tick frequency.

But this has nothing to do with the harness. The precision of the timer would naturally vary across platforms. That is why there is the extra wrapper so we can easily substitute for the most precise time source according to the platform (e.g. HPET on Intel etc.)


> Now, in my opinion, we need an implementation that runs the benchmark loop until a particular time occurs, i.e. it counts the number of iterations taken before a particular kernel tick.

Can you sketch a pseudo-code because I really do not understand what you mean. It seems to me that you suggest that the benchmark should check whether enough time passed and then terminate. That does not differ much from the current concept except it brings more overhead of the timer.

> This naturally results in more complex code, as well as some overhead from timer checking, which needs to be accounted for, but would give vastly better results and usability.

Are you basing your observations (accounting for timer overhead + vastly better results) on intuition or on some factual results? I would love to see the data supporting your arguments. 

And if usability means shorter execution time: are we interested in one-shot performance then? I assume we are more interested in stable-state performance of the servers than short-lived applications (e.g. VFS vs. `ls`).

le-jzr:
It's not just a matter of time taken, but the economy of its use. Ten samples of ten seconds give you total runtime of 100s. Taking 100s to gather ten samples and compute standard deviation of those ten samples strikes me as useless. Why not take 100x1s samples? Or 1000x0.1s samples? Doing that would give you more and better information, but we can't, because we depend on granularity of the timer, which can be as high as 0.01 or 0.1s, in absence of more precise hardware timers. Such data would be meaningless. The answer is to measure in a way that's independent of the timer granularity.

But you're right, I haven't presented any data showing the impact, nor code. I'll try to remedy that.

Dzejrou:
> > Now, in my opinion, we need an implementation that runs the benchmark loop until a particular time occurs, i.e. it counts the number of iterations taken before a particular kernel tick.
> 
> Can you sketch a pseudo-code because I really do not understand what you mean. It seems to me that you suggest that the benchmark should check whether enough time passed and then terminate. That does not differ much from the current concept except it brings more overhead of the timer.

Wouldn't this also require you to know beforehand what the approximate duration of the task will be? If you specify an absolute amount of time for a benchmark to run, the number of iterations performed may vary (even greatly) depending on the type of task measured. It changes the question of 'How many times do I want to repeat a task to get meaningful results?' to 'How much time would I need to give this for it to run the task enough times to get meaningful results?', which adds quite a bit of cognitive load IMHO.

Additionally, the time a certain task takes may possibly change between versions and as such a version change of the task may alter the meaningfulness of the measurements. Or am I missing something?

vhotspur:
Hopefully last bunch of commits. Summary of changes follows, I believe it can be merged in the current state.

* Removed the global hash table, turned it into an extra parameter of each benchmark (practically, the parameters *were* global but it is probably cleaner to do it this way).
* Simplified error reporting from within a benchmark runner.
* Add options to specify minimal duration of a loop and number of runs to execute.

There are still some bike-shedding issues.

I believe the functions to start/stop the timer should be in-lined to minimize performance overhead (there are currently two wrappers so the impact could be there). I have not measured the difference but I think it cannot hurt anything (the compiler may decide not to in-line, after all).

Name of the whole beast. I renamed it to `hbench` as it is a set of benchmarks for HelenOS (and the name is easy to type and describes well what the application does). I do not like when commands are named too generally as it complicates searching. For example, *download* is mentioned almost 40 times on our wiki, but only once when talking about our tool. Hence the `h` prefix. Since it contains HelenOS-specific benchmarks, I do not think (a) it hurts anything to include it in the name, (b) it will be ever ported to a different platform and even then, what is wrong if the benchmark still contains the original name?

Anyway, unless there are some serious blockers, I would like to merge this in its current state. I believe it improves over the current state (code deduplication, merge with another similar application, new benchmark, more documentation) and there is plenty of time before next release if something would need to be changed.

PR #153, opened 2019-01-11
Remove async "fast" and "slow" functions from public interface

le-jzr:
"fast" and "slow" paths are implementation detail.
All macros in <async.h> are turned into external functions, so that
this implementation detail doesn't leak. Additionally, removing macros is
A Good Thing on its own, e.g. helping C++ interoperability.

### DIFF
--- a/uspace/lib/c/include/async.h
+++ b/uspace/lib/c/include/async.h
@@ -107,34 +107,18 @@ typedef struct async_exch async_exch_t;
 
 extern __noreturn void async_manager(void);
 
-#define async_get_call(data) \
-	async_get_call_timeout(data, 0)
-
+extern bool async_get_call(ipc_call_t *);
 extern bool async_get_call_timeout(ipc_call_t *, usec_t);
 
-/*
- * User-friendly wrappers for async_send_fast() and async_send_slow(). The
- * macros are in the form async_send_m(), where m denotes the number of payload
- * arguments. Each macros chooses between the fast and the slow version based
- * on m.
- */
-
-#define async_send_0(exch, method, dataptr) \
-	async_send_fast(exch, method, 0, 0, 0, 0, dataptr)
-#define async_send_1(exch, method, arg1, dataptr) \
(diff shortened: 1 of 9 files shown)


jxsvoboda:
What is the performance impact?

uspace/lib/c/include/async.h
```diff
@@ -356,93 +296,32 @@ extern void async_sess_args_set(async_sess_t *, iface_t, sysarg_t, sysarg_t);
...
 
 extern errno_t async_share_out_start(async_exch_t *, void *, unsigned int);
 extern bool async_share_out_receive(ipc_call_t *, size_t *, unsigned int *);
 extern errno_t async_share_out_finalize(ipc_call_t *, void **);
 
-/*
- * User-friendly wrappers for async_data_read_forward_fast().
- */
-
-#define async_data_read_forward_0_0(exch, method, answer) \
```
jermar: So where is async_data_read_forward_x_y() and async_data_write_forward_x_y() for 0 <= x < 4 && 0 <= y <= 1?

le-jzr: Good question.
I was thinking that since it's such a fringe use function, we don't really need all that API clutter for it (note those macros are currently not used anywhere). I also wanted to get rid of the `_fast` suffix. So I added just the `_4_1` variant thinking we could add more variants later if needed.

le-jzr:
> What is the performance impact?

I'd expect the extra indirection to be insignificant relative to the IPC operation itself, but I'll investigate.

uspace/lib/c/include/async.h
le-jzr: Yes, it does seem arbitrary. No, it wouldn't be much work. I'll add them.

le-jzr:
@jxsvoboda I did just _very_ rudimentary tests using `perf ns_ping` and `perf ping_pong`, but the results I see are pretty much identical. That's in line with my belief that there is no reason for it to affect performance.

PR #154, opened 2019-01-12
Abort and report when an unimplemented function is called

Dzejrou:
As discussed with @le-jzr , I added a macro that is used in unimplemented functions to abort the program and report the source code location.

There were some cases ignored:
  - rbtree: the missing functions are meant to be no-op (it's missing balancing operations, but the
                structure works without them)
  - locale: these functions modify their input to another localization (e.g. encoding), but they are stubs
  - constexpr functions: these can't be done without implementation or removal, there were two cases:
    - string.hpp: does not matter, their return value is valid
    - limits.hpp: 3 floating point functions, I will look for a fix (or remove them)

Also added missing include in <new> and a missing symbol needed for the runtime (it was part of the custom linker scripts which are now gone).


le-jzr:
Awesome, thanks for this!

PR #155, opened 2019-02-01
Avoid directly using .head/.next/.prev of list_t/link_t

le-jzr:
Use existing constructs from <adt/list.h> instead.

---

Probably needs to be reviewed a few more times.

Apart from making better use of the existing abstraction, this could also allow possibly changing some painful aspects of the lists, in particular their self-referential initialization. This is a personal opinion, but I feel like they should work properly with just zero initialization. Either way, avoiding direct use of member pointers makes it possible to experiment with alternatives.

### DIFF
--- a/uspace/lib/c/generic/adt/hash_table.c
+++ b/uspace/lib/c/generic/adt/hash_table.c
@@ -272,14 +272,14 @@ hash_table_find_next(const hash_table_t *h, ht_link_t *first, ht_link_t *item)
 	assert(h && h->bucket);
 
 	size_t idx = h->op->hash(item) % h->bucket_cnt;
+	list_t *list = &h->bucket[idx];
 
 	/* Traverse the circular list until we reach the starting item again. */
-	for (link_t *cur = item->link.next; cur != &first->link;
-	    cur = cur->next) {
-		assert(cur);
+	for (link_t *cur = list_next(&item->link, list); cur != &first->link;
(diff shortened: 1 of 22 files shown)


uspace/lib/c/generic/adt/hash_table.c
```diff
@@ -272,14 +272,14 @@ hash_table_find_next(const hash_table_t *h, ht_link_t *first, ht_link_t *item)
...
-	for (link_t *cur = item->link.next; cur != &first->link;
-	    cur = cur->next) {
-		assert(cur);
+	for (link_t *cur = list_next(&item->link, list); cur != &first->link;
+	    cur = list_next(cur, list)) {
 
-		if (cur == &h->bucket[idx].head)
-			continue;
+		if (!cur)
+			cur = list_first(list);
```
jermar: This does not seem right. If 'first' is literally the first item on the list, the loop does not terminate on it (as it should) and the loop can create a livelock together with irq_dispatch_and_lock_table().

jermar:
Besides the one issue I found, the rest LGTM. Please test thoroughly though. I believe it might even make sense to break this into a series of smaller commits.

jxsvoboda:
I don't have an issue with removing the cases where we access the fields directly and I've looked at it and also seems OK to me except the one issue found by Jakub. On the other hand I don't like the way you used to prevent the field access in the future making the field names ugly. I suggest we treat both issues separately.

PR #156, opened 2019-02-01
Change the meaning of `counter` in fibril mutex

le-jzr:
Previously analogous to semaphore tokens, initially 1 and
counting downward into negative numbers.

Now the number of fibrils currently accessing or waiting for access
to the critical section. Initially 0 (unlocked mutex), and counting
up. Never negative when used correctly. Also consistent with
readers/writers counts in rwlock.

This way is IMO easier to understand.

### DIFF
--- a/uspace/lib/c/generic/thread/fibril_synch.c
+++ b/uspace/lib/c/generic/thread/fibril_synch.c
@@ -164,7 +164,7 @@ static void check_for_deadlock(fibril_owner_info_t *oi)
 void fibril_mutex_initialize(fibril_mutex_t *fm)
 {
 	fm->oi.owned_by = NULL;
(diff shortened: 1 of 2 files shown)


le-jzr:
Well, if we additionally get rid of the list self-reference, zeroed out memory would be a valid initial state for the mutex, which would mean a few bytes moved from `.data` to `.bss`. So, not really, no.

le-jzr:
There's missing context here. This came about when I was figuring out the issue of libcpp standard headers transitively including `<adt/list.h>`. At first I solved it by not including that and instead defining the mutex as few opaque words of data for which zeroed out would have been valid initial state. Later I made `<adt/list.h>` itself C++-aware, so this is not really helpful any more.

PR #157, opened 2019-02-02
Turn some function-like macros into functions

le-jzr:
This is part of a more extensive effort to reduce use of preprocessor defines, in order for our headers to play nicer with code analysis tools and with C++.

Note that it's somewhat easier to review the changes commit by commit, than it is to review the whole pull request in one diff.

### DIFF
--- a/abi/include/abi/ipc/ipc.h
+++ b/abi/include/abi/ipc/ipc.h
@@ -35,8 +35,10 @@
 #ifndef ABI_IPC_IPC_H_
 #define ABI_IPC_IPC_H_
 
+#include <stdint.h>
 #include <abi/proc/task.h>
 #include <abi/cap.h>
+#include <_bits/errno.h>
 
 /** Length of data being transferred with IPC call
  *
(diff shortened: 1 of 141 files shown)


abi/include/abi/ipc/ipc.h
```diff
@@ -72,22 +72,22 @@
...
-
-#define IPC_GET_IMETHOD(data)  ((data).args[0])
-#define IPC_GET_RETVAL(data)   ((errno_t) (data).args[0])
-
-#define IPC_GET_ARG1(data)  ((data).args[1])
-#define IPC_GET_ARG2(data)  ((data).args[2])
-#define IPC_GET_ARG3(data)  ((data).args[3])
-#define IPC_GET_ARG4(data)  ((data).args[4])
-#define IPC_GET_ARG5(data)  ((data).args[5])
+#define IPC_SET_RETVAL(data, retval)  ((data)->args[0] = (sysarg_t) (retval))
```
jermar: This seems to be changing one way of doing things with another, without changing anything substantial. There are still macros, we only add & or remove * in some places in order to be able to use -> inside of the macro. Is there something more beyond this huge commit?

abi/include/abi/ipc/ipc.h
```diff
@@ -127,6 +111,77 @@ typedef struct {
 	cap_call_handle_t cap_handle;
 } ipc_data_t;
 
+/* Functions for manipulating calling data */
+
+static inline void ipc_set_retval(ipc_data_t *data, errno_t retval)
```
jermar: Ah ok, this makes sense.

PR #159, opened 2019-02-05
Avoid using ambiguous terminology in str.h/str.c

le-jzr:
Instead of ambiguous terms "size", "length", and "characters", use unambiguous terms "bytes" and "code points", including in function names.

Function name changes:
`(w)str_(ln)size` -> `(w)str_(ln)bytes`
`(w)str_(n)length` -> `(w)str_(n)code_points`

The purpose of this change is to avoid confusing the purpose of those functions, especially given the conflicting terminology of standard `strlen`, and the nonexistence of the term "character" in Unicode (there we have code units, code points, and grapheme clusters, only the last of which actually being what's colloquially understood as a character).

### DIFF
--- a/boot/generic/include/str.h
+++ b/boot/generic/include/str.h
@@ -47,9 +47,9 @@
 extern wchar_t str_decode(const char *str, size_t *offset, size_t sz);
 extern errno_t chr_encode(wchar_t ch, char *str, size_t *offset, size_t sz);
 
-extern size_t str_size(const char *str);
-extern size_t str_lsize(const char *str, size_t max_len);
-extern size_t str_length(const char *str);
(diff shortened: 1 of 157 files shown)


le-jzr:
@jermar, @jxsvoboda Any comments?

jermar:
Hm, I don't feel knowledgeable enough, so whatever you and @jxsvoboda deem proper.

jxsvoboda:
I understand where you are coming from, but you will notice the cunning correspondence between the 'l' prefix and the 'length' suffix (although not for n vs size which gets lost with your change. Also _code_points is wordier than _length. Overall I am not convinced this change is for the better.

PR #160, opened 2019-02-12
Uspace pcut

matthieuriolo:
Some bugfixes/improvements and new unit tests for cap, uuid, getopt, gsort, double_to_str and ieee_double. Some of the tests fail because the implementation does not behave as expected.

### Todo
cap.c fails when trying to convert 999999. Expected 1.000 MB but get 1000.0 kB. Appears that it fails to increase the unit when the number gets rounded.
The following failure appears only on ia32 but not on amd64, it fails for 555999999 which should convert to 5.56 GB but instead 1.261 GB is returned.

getopt.c fails when reading an existing but optional parameter. See PCUT_TEST(getopt_optional_param). ~~It also fails when there are two flags separated by a whitespace, see PCUT_TEST(getopt_flag_whitespace). It works though when it is a option with a parameter, see the PCUT_TEST(getopt_param_whitespace)~~

ieee_double probably fails on Big Endian Systems and will effect the output of printf("%f"). This needs to be tested and is not verfied yet.

### Improvements
The files ieee_double and double_to_str are only used in printf_core. Maybe they should be located somewhere else than in the root of libc 

When writing the gsort test and I came across a bug which made me think of a solution how it could have prevented in the first place:

1. Create a new type for the comparator function. Shorter syntax and more readable
2. Make this new type return a meaningful type instead of an int. Something like [cfcomparisonresult](https://developer.apple.com/documentation/corefoundation/cfcomparisonresult)
3. relocate qsort, gsort and bubblesort (from test/qsort) to lib/c/sort/
and create a wrapper function qsort in the root of libc


uspace/lib/c/generic/ieee_double.c
```diff
@@ -91,7 +91,12 @@ ieee_double_t extract_ieee_double(double val)
 		/* Denormal or zero. */
 		if (ret.is_denormal) {
 			ret.pos_val.significand = raw_significand;
-			ret.pos_val.exponent = 1 - exponent_bias;
+			if (raw_significand == 0) {
+				ret.pos_val.exponent = -exponent_bias;
+			} else {
+				ret.pos_val.exponent = 1 - exponent_bias;
+			}
+
```
le-jzr: This change seems meaningless to me. If significand is 0, the value of exponent doesn't matter.

uspace/lib/c/test/main.c
```diff
@@ -40,19 +40,19 @@ PCUT_IMPORT(odict);
...
 PCUT_IMPORT(string);
 PCUT_IMPORT(table);
+PCUT_IMPORT(imath);
 PCUT_IMPORT(cap);
 PCUT_IMPORT(gsort);
-PCUT_IMPORT(ieee_double);
-PCUT_IMPORT(double_to_str);
 PCUT_IMPORT(getopt);
 PCUT_IMPORT(uuid);
-PCUT_IMPORT(imath);
```
le-jzr: Could your commit message explain why the order was wrong, and why this change fixes it?

uspace/lib/c/test/main.c
```diff
@@ -40,12 +40,19 @@ PCUT_IMPORT(odict);
...
 PCUT_IMPORT(stdio);
 PCUT_IMPORT(stdlib);
 PCUT_IMPORT(str);
 PCUT_IMPORT(string);
 PCUT_IMPORT(table);
+PCUT_IMPORT(imath);
+PCUT_IMPORT(cap);
+PCUT_IMPORT(gsort);
+PCUT_IMPORT(getopt);
+PCUT_IMPORT(uuid);
```
le-jzr: The imports should be listed in lexicographic order.

uspace/lib/c/generic/ieee_double.c
matthieuriolo: 0 is a special case (like infinity and NaN). If exponent != 0 then the hidden bits exist. If exponent == 0, then no hidden bits. Keep in mind that the exponent is underflowed in our case

le-jzr: This already is the denormal/zero case. The output is a representation where no bits are hidden.

le-jzr: What I mean is, the semantics of `ret` are that the actual numeric value of the number is `pos_val.significand * 2^pos_val.exponent`. If `raw_significand == ret.pos_val.significand == 0`, then `pos_val.exponent` can be arbitrary.

matthieuriolo: but then you have a discrepant between the flag is_denormal and the exponent

le-jzr: The flags describes the original IEEE 754 format, it does not affect the semantics of the converted value.

matthieuriolo: I tested it accordingly to the IEEE specification and it's states that the exponent must be 0. I dont think that there is any room for argumentation, either it behaves accordingly to the specification or it doesnt. If you think behaving "similiar" is good enough, fine. But it's wrong.

le-jzr: I was under the impression that `ieee_double_t` is purely our own thing. Is the structure and its meaning specified by the IEEE standard?

matthieuriolo: Really? In that case we better call it `helenos_something_nummeric_t`

le-jzr: Forgive me if I'm being obtuse, but I truly don't understand what you are talking about. You are saying `0.0` should be represented by `(fp_num_t) { .significand = 0, .exponent = -1075 }`, instead of `(fp_num_t) { .significand = 0, .exponent = -1074 }`. You'll have to explain to me why that matters, because I don't understand.

le-jzr: So that you don't misunderstand my intent, please understand that having "standard-looking" interfaces which are not behaving according to standard is something we are very much trying to obliterate from our code base. If you found such a thing, fixing it is very much wanted and appreciated. That being said, I don't understand which part of the standard would be relevant to this `(integer, binary exponent)` tuple representation. As far as I know, no part of the standard even acknowledges existence of such representations, despite their inherent usefulness in math code.

matthieuriolo: > You'll have to explain to me why that matters, because I don't understand.

Because that's not what you would expect to get when you pass `0.0` to that function. The function clearly says it represents the IEEE standard, and that one specifies that  `0.0` => exponent = significant = 0

Or let me rephrase it, why do you think is `-1074` a better choice? Because it's a random number? Perfomance boost?

le-jzr: I don't think `-1074` is a better choice. I think adding extra branches just to prefer `-1075` in the place of a number which has no meaning is pointless. IEEE standard specifies the formatting of the _input_ of this function. As I said above, the output is an `(integer, exponent)` tuple, and you are proposing we should specifically prefer `-1075` as the exponent in that tuple.

le-jzr: Why would you even expect a specific value there for 0? If it makes difference for your code, the code is just written wrong.

matthieuriolo: In all fairness, I've thought about that too. It collides with other functions though => which is exactly the reason I dont like it to be "random"/undocumented and prefer an extra test to ensure you get an expected result. I can revert that commit, but then you can not meaningful test for the exponent in tests and you must hope that you never have to touch any of the depending functions since you will be left with only one documentation and that will be the source code.

le-jzr: If you just want a single concrete value for the tuple, why is `-1075` any better than `-1074`?
It would make sense if you wanted to set it to `0`. Meaningless, but sensible. But that's not what you are proposing.

le-jzr: I think you might be confusing the _actual_ exponent with its IEEE _representation_. `double` contains IEEE _representation_ of the exponent, in which all-zero exponent field has a special meaning which could be misinterpreted by applying the bias used for normalized numbers. However, the `fp_pos_t` type used here is the tuple (pair) of `(integer, exponent)`, both of which are their pure mathematical, unbiased value. It has nothing to do with the formats specified by IEEE 754, it's a different format, which is useful for doing math on the value.

matthieuriolo: well I guess that's fair

le-jzr:
Nice, thanks! 
I haven't had much time to review today, but I'll make sure to review and merge your changes by the end of the week, unless someone beats me to it.

uspace/lib/c/generic/ieee_double.c
matthieuriolo: Reverted the commit and corrected tests accordingly (hence remove the test of the exponent in case of `0.0`)

uspace/lib/c/generic/uuid.c
```diff
@@ -64,8 +64,8 @@ errno_t uuid_generate(uuid_t *uuid)
 	}
 
 	/* Version 4 UUID from random or pseudo-random numbers */
-	uuid->b[8] = (uuid->b[8] & ~0xc0) | 0x40;
-	uuid->b[6] = (uuid->b[6] & 0xf0) | 0x40;
+	uuid->b[6] = (uuid->b[6] & 0x4F) | 0x40;
+	uuid->b[8] = (uuid->b[8] & 0xBF) | 0xB0;
```
le-jzr: This is incorrect. I committed a proper fix to the mainline.

le-jzr: To elaborate, `| 0xB0` is setting more bits than it should. The rest is ok, but my change also has slightly different `&` masks as a matter of personal preference.

matthieuriolo: You're right, the `| 0xB0` part would end up making it exactly B - which of course is wrong since it can be 8, 9, A or B. It supposed to be a fallback when `& 0xB` ends up being `0x0`. It should be therefore `| 0x80`

jermar:
Or don't rebase / merge at all unless you really need to rebase.

matthieuriolo:
Should I remove the last commit therefore?

jermar:
Yes, please.

uspace/lib/c/generic/cap.c
```diff
@@ -159,7 +159,9 @@ void cap_simplify(cap_spec_t *cap)
 	}
 
 	/* Round the number so that we have at most @c scap_max_sdig significant digits */
-	sdig = 1 + ilog10_u64(cap->m); /* number of significant digits */
+	rc = ilog10_u64(cap->m, &sdig); /* number of significant digits */
+	sdig += 1;
+	assert(rc == EOK);
```
le-jzr: I don't think we can assume that `cap->m` is never zero.

le-jzr:
Hi, sorry about my radio silence.
@matthieuriolo I rebased and reorganized your changes a little, see [this branch](https://github.com/le-jzr/helenos/commits/pcut_tests). If you have no objections, I'd merge that rebased branch. I omitted "make ilog10_u64() return an errno_t" for now, because of my comment above.

uspace/lib/c/generic/cap.c
matthieuriolo: Your right again. But this means I also forgot to implement a test for caps where 0 gets passed. If not I should have noticed it, since it always should fail when `cap->m == 0`. I assume you agree that `ilog10_u64(0)` should indeed fail with an errno?

matthieuriolo:
that's fine by me. Thx for your review and efforts.

I do wonder though about the existing fails and my proposes from the pull request. What happened to them?

le-jzr:
Well, I don't really feel like digging into the failures right now. As far as I'm concerned, failing tests are better than no tests, and someone will fix them eventually.

The suggestion for comparison functions has merit, but I don't think an enum works. Testing for e.g. "less then or equal" would get pretty verbose. It would make sense to retain the current integer value, but hidden inside a struct, which would then only be tested via some functions provided by the library. The code would then look something like `if (islesseq(compare(x, y))`. Feel free to work on it if you want, I'm certainly very interested in any changes that reduce the likelihood of mistakes. :)

le-jzr:
I merged the rebased version of this pull request I mentioned above. I think the only missing part is the `ilog10_u64` change. Let me know if anything's amiss, otherwise I think you can close this PR.

le-jzr:
Oh, and thanks a lot! :)

matthieuriolo:
> As far as I'm concerned, failing tests are better than no tests, and someone will fix them eventually.

I will therefore give it another try. Well, maybe... I'm currently working on something else

>  Testing for e.g. "less then or equal" would get pretty verbose. It would make sense to retain the current integer value, but hidden inside a struct, which would then only be tested via some functions provided by the library. The code would then look something like `if (islesseq(compare(x, y))`

Mh, are you not overshooting? `islesseq()` is the same as `!isgreater()`. And this can be done with an enum and the `==`/`!=` comparator , at least I thought so. Again, maybe I will give it a try when I've time for it. I simply wanted to have a validation that this indeed is a good idea before I invest time for something which will be rejected anyway.

>  I think the only missing part is the `ilog10_u64` change. Let me know if anything's amiss, otherwise I think you can close this PR

I actually changed my mind on `ilog10_u64`. It probably would be better to define in the documentation that it returns 0 instead of an error code. But yes, I think **everything is included and this pull request can be closed**

> Oh, and thanks a lot! :)

You're welcome :)

le-jzr:
> Mh, are you not overshooting? `islesseq()` is the same as `!isgreater()`.

Good point, you're right. :)
Another think that might potentially enter consideration is that using enum would mean extra branches where previously there would be just a subtraction, which could make a difference for performace. Though that's just a suspicion, I haven't actually tested any of that.

matthieuriolo:
> Another thing that might potentially enter consideration is that using enum would mean extra branches where previously there would be just a subtraction, which could make a difference for performace

This is indeed true - If I remember correctly an IF consumes as many ticks as an SUBTRACT. You are indeed on something here.

PR #161, opened 2019-02-13
C++ compatibility improvements

le-jzr:


abi/include/_bits/errno.h
```diff
@@ -41,9 +41,9 @@
 #define _BITS_ERRNO_H_
 
 #ifdef __OPAQUE_ERRNO__
-#include <_bits/opaque_handle.h>
+#include <_bits/__opaque_handle.h>
```
Dzejrou: It seems to me you changed the include directive but didn't actually rename the file _bits/opaque_handle.h to _bits/__opaque_handle.h, or am I missing something?

le-jzr: You're right, good catch!

uspace/lib/cpp/include/__bits/limits.hpp
```diff
@@ -527,7 +527,7 @@ namespace std
 
             static constexpr unsigned short min()
             {
-                return USHRT_MIN;
+                return 0;
```
Dzejrou: No problem with these changes, but don't they trigger warnings about implicit signed -> unsigned conversion?

uspace/lib/cpp/include/__bits/locale/num_get.hpp
```diff
@@ -300,18 +300,20 @@ namespace std
...
+                        res = ::strtoll(base.buffer_, nullptr, num_base);
                     else
-                        ret = std::hel::str_uint64_t(base.buffer_, nullptr, num_base, false, &res);
+                        res = ::strtoull(base.buffer_, nullptr, num_base);
 
-                    if (ret != EOK)
-                    {
+                    if (errno != EOK)
                         err |= ios_base::failbit;
-                        v = 0;
```
Dzejrou: The zero assignment here is required by $22.4.2.1.2 (n4296 [1] bottom of page 717, point (3.4)). This is because if we use something like:

    int x = 3;
    std::cin >> x;

And we input something that isn't convertible, the value of `x` should be `0`, not `3`.

Dzejrou: [1] http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4296.pdf

uspace/lib/cpp/include/__bits/limits.hpp
le-jzr: They don't, as far as I know.

uspace/lib/cpp/include/__bits/locale/num_get.hpp
le-jzr: `x` is assigned later. The original behavior wouldn't be consistent with the standard, see behavior of `strtoll()`:
>If the value read is out of the range of representable values by a long long int, the function returns LLONG_MAX or LLONG_MIN (defined in <climits>), and errno is set to ERANGE.

le-jzr: Actually, now that I'm looking at it, the C function doesn't guarantee `errno` being set at all when there are no digits in the input. This needs an extra check.

le-jzr: So I dug a little deeper and discovered a bunch of problems in implementation of both the standard and nonstandard conversion functions. I suppose it's as good a time as any to finally start making pcut tests...

Dzejrou:
Hi, any plans to merge?

le-jzr:
Yes. I really want to get back to it this week.

Dzejrou:
Great, thank you!

le-jzr:
Looks like my long period of inactivity didn't introduce any conflicts, thankfully. I opened pull request #169 with fixes for libc functions on which part of libcpp depends. I still have to check that nothing here regresses coastline.

le-jzr:
It breaks gcc build for some reason. :-/ Gotta fix that.

le-jzr:
Seems to be working. Let's see what CI says.

PR #162, opened 2019-02-14
[RFC] Do not use preprocessor macros in `abi/`

le-jzr:
Builds upon PR#157, converting most of the macros present in `abi/` to C language constructs.


jermar:
Why not. The biggest issue I have with this is that certain things are no longer in capital letters, but that's just a matter of getting used to.

PR #163, opened 2019-03-02
Ticket 684 improved

matthieuriolo:
Based on #147 

- Allows to execute subdirectories from search_path and to autocomplete them
- prevents execution from the working directory and autocompletion if not specified with ./ ../ or /

### DIFF
--- a/uspace/app/bdsh/compl.c
+++ b/uspace/app/bdsh/compl.c
@@ -39,6 +39,7 @@
 #include "compl.h"
 #include "exec.h"
 #include "tok.h"
+#include "util.h"
 
 static errno_t compl_init(wchar_t *text, size_t pos, size_t *cstart, void **state);
 static errno_t compl_get_next(void *state, char **compl);
@@ -208,15 +209,24 @@ static errno_t compl_init(wchar_t *text, size_t pos, size_t *cstart, void **stat
 			goto error;
 		}
 		*cstart += rpath_sep + 1 - prefix;
-		free(prefix);
-		prefix = NULL;
 
 		cs->path_list = malloc(sizeof(char *) * 2);
 		if (cs->path_list == NULL) {
 			retval = ENOMEM;
 			goto error;
 		}
-		cs->path_list[0] = dirname;
+
+		if (!is_path(prefix) && cs->is_command) {
+			cs->path_list[0] = malloc(sizeof(char) * PATH_MAX);
+			int ret = snprintf(cs->path_list[0], PATH_MAX, "%s/%s", search_dir[0], dirname);
+			if (ret < 0 || ret >= PATH_MAX) {
+				retval = ENOMEM;
+				goto error;
+			}
+		} else {
+			cs->path_list[0] = dirname;
+		}
+
 		cs->path_list[1] = NULL;
 		/*
 		 * The second const ensures that we can't assign a const
--- a/uspace/app/bdsh/compl.c
+++ b/uspace/app/bdsh/compl.c
@@ -218,6 +218,11 @@ static errno_t compl_init(wchar_t *text, size_t pos, size_t *cstart, void **stat
 
 		if (!is_path(prefix) && cs->is_command) {
 			cs->path_list[0] = malloc(sizeof(char) * PATH_MAX);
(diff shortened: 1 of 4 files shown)


uspace/app/bdsh/compl.c
```diff
@@ -208,15 +209,24 @@ static errno_t compl_init(wchar_t *text, size_t pos, size_t *cstart, void **stat
 			goto error;
 		}
 		*cstart += rpath_sep + 1 - prefix;
-		free(prefix);
```
jermar: So can prefix be leaked now?

matthieuriolo: Good call, it's fixed now

matthieuriolo:
Since bdsh always could execute subdirectories in the search_path (but not autocomplete them) the commit 31872f7e7a30e53e2e8f02bd6b1c3eb254af29a9 can be reverted

uspace/app/bdsh/compl.c
```diff
@@ -208,15 +209,24 @@ static errno_t compl_init(wchar_t *text, size_t pos, size_t *cstart, void **stat
...
 
 		cs->path_list = malloc(sizeof(char *) * 2);
 		if (cs->path_list == NULL) {
 			retval = ENOMEM;
 			goto error;
 		}
-		cs->path_list[0] = dirname;
+
+		if (!is_path(prefix) && cs->is_command) {
+			cs->path_list[0] = malloc(sizeof(char) * PATH_MAX);
```
jermar: Return value from malloc is not tested here. I think C99 snprintf() reacts gracefully to a NULL str, but I am not sure this is also the case of our snprintf. Better test the value returned from malloc() to be on the safe side.

uspace/app/bdsh/exec.c
```diff
@@ -73,26 +70,27 @@ static int try_access(const char *f)
...
 	/* The user has specified a full or relative path, just give it back. */
-	if (-1 != try_access(cmd)) {
-		return (char *) cmd;
+	if (is_path(cmd)) {
+		if (-1 != try_access(cmd)) {
+			return str_dup(cmd);
+		}
 	}
 
+	char *found = (char *)malloc(PATH_MAX);
```
jermar: Again, found can be NULL. Also the case of original code, but this should be fixed.

uspace/app/bdsh/exec.c
```diff
@@ -106,8 +104,7 @@ unsigned int try_exec(char *cmd, char **argv, iostate_t *io)
 	int file_handles[3] = { -1, -1, -1 };
 	FILE *files[3];
 
-	tmp = str_dup(find_command(cmd));
-	free(found);
+	tmp = find_command(cmd);
```
jermar: If this is NULL, we should return error here.

uspace/app/bdsh/util.c
```diff
@@ -73,3 +73,16 @@ unsigned int cli_set_prompt(cliuser_t *usr)
 
 	return 0;
 }
+
+/*
+ * Returns true if the string is a relative or an absolute path
+ */
+bool is_path(const char *cmd)
```
jermar: This does not work for relative paths that do not start with './'. So for example, if you copy tetris to /foo/bar/qqq and then cd to /foo, you can run this copy of tetris by typing bar/qqq, but bdsh will still not autocomplete it for you.

matthieuriolo: `you can run this copy of tetris by typing bar/qqq, but bdsh will still not autocomplete it for you`

I cannot reproduce your complain. [Screenshot](https://prnt.sc/mynkes)

When using _without_ ./, ../ or / it neither autocompletes nor executes it from the workingdirectory - which I would deem as the correct behaviour

jermar: Works for me. Are we on the same revision? Your screenshot says b295720, while mine is 1c481ee.

matthieuriolo: Alright, some silly mistakes on my sides. I was not aware that `make clean` does not reset the revision number. I actually solved this issue already, but I messed up to force push it earlier. The new commit contains it accidentely...

PR #165, opened 2019-04-06
Add support for AArch64

setupji:
This patch series adds support for the 64-bit Arm architecture (AArch64) to HelenOS. It targets the QEMU virt platform with Cortex-A57, PL011 (UART), GICv2 (interrupt controller), VirtIO-SCSI CD-ROM and EDK2 (UEFI implementation). The port implements usual low-level bits needed to run HelenOS on some architecture (boot, kernel, uspace) and can successfully boot on the mentioned platform and run user space console applications.

Documentation:
* ARM® Architecture Reference Manual -- ARMv8, for ARMv8-A architecture profile
  https://static.docs.arm.com/ddi0487/da/DDI0487D_a_armv8_arm.pdf
* ARM® Generic Interrupt Controller -- Architecture version 2.0 -- Architecture Specification
  https://static.docs.arm.com/ihi0048/bb/IHI0048B_b_gic_architecture_specification.pdf
* PrimeCell® UART (PL011) -- Revision: r1p5 -- Technical Reference Manual
  https://static.docs.arm.com/ddi0183/g/DDI0183G_uart_pl011_r1p5_trm.pdf


boot/arch/arm64/src/main.c
```diff
@@ -0,0 +1,328 @@
...
+		goto fail;
+	}
+
+	/*
+	 * Check that everything is aligned on a 4kB boundary and the kernel can
+	 * be placed by the decompression code at a correct address.
+	 */
+
+	/* Statically check PAGE_SIZE and BOOT_OFFSET. */
+#if PAGE_SIZE != 4096
```
jermar: Can also use _Static_assert for this

kernel/genarch/src/mm/as_pt.c
```diff
@@ -75,9 +75,9 @@ pte_t *ptl0_create(unsigned int flags)
 	pte_t *dst_ptl0 = (pte_t *)
 	    PA2KA(frame_alloc(PTL0_FRAMES, FRAME_LOWMEM, PTL0_SIZE - 1));
 
-	if (flags & FLAG_AS_KERNEL)
-		memsetb(dst_ptl0, PTL0_SIZE, 0);
-	else {
+	memsetb(dst_ptl0, PTL0_SIZE, 0);
+
+	if (!KERNEL_SEPARATE_PTL0 && !(flags & FLAG_AS_KERNEL)) {
```
jermar: So what's the deal with KERNEL_SEPARATE_PTL0, why is this done differently on arm64?

tools/ew.py
```diff
@@ -93,6 +93,22 @@ def platform_to_qemu_options(platform, machine, processor):
 		return 'system-x86_64', pc_options(64)
 	elif platform == 'arm32':
 		return 'system-arm', '-M integratorcp'
+	elif platform == 'arm64':
+		# Check that ROM image is present. Provide the user with
+		# appropriate steps to fix this problem.
```
jermar: It would be nice to support placing the image in some location where QEMU stores other firmware images, so that it does not have to be in the HelenOS root.

kernel/generic/include/mm/as.h
```diff
@@ -58,6 +58,12 @@
  */
 #define KERNEL_ADDRESS_SPACE_SHADOWED  KERNEL_ADDRESS_SPACE_SHADOWED_ARCH
 
+/**
+ * Defined to be true if user address space and kernel address space do not
+ * share the same page table.
+ */
+#define KERNEL_SEPARATE_PTL0 KERNEL_SEPARATE_PTL0_ARCH
```
jermar: I see

uspace/srv/hid/input/input.c
```diff
@@ -628,6 +628,9 @@ static void kbd_add_legacy_devs(void)
 #endif
 #if defined(UARCH_sparc64) && defined(PROCESSOR_sun4v)
 	kbd_add_dev(&chardev_port, &stty_ctl);
+#endif
+#if defined(UARCH_arm64) && defined(MACHINE_virt)
+	kbd_add_dev(&chardev_port, &stty_ctl);
```
jermar: We should probably do this by passing the 'console' boot argument as we now do on other platforms. However, let's do this after the merge.

jermar:
LGTM.
Let's wait a couple of days to allow others to review this too.

kernel/arch/arm64/src/interrupt.c
```diff
@@ -0,0 +1,173 @@
...
+ *
+ * @return Old interrupt priority level.
+ */
+ipl_t interrupts_disable(void)
+{
+	uint64_t daif = DAIF_read();
+
+	DAIF_write(daif | DAIF_IRQ_FLAG);
+
+	return (daif >> DAIF_IRQ_SHIFT) & 1;
```
le-jzr: It would look nicer if `ipl_t` was simply `(daif & DAIF_IRQ_FLAG)`.

kernel/arch/arm64/src/start.S
```diff
@@ -0,0 +1,288 @@
...
+
+#include <abi/asmtool.h>
+#include <arch/boot/boot.h>
+#include <arch/mm/km.h>
+#include <arch/mm/page.h>
+#include <arch/regutils.h>
+
+.section K_TEXT_START, "ax"
+
+SYMBOL(kernel_image_start)
```
le-jzr: Is it necessary for this whole function to be written in assembly? I feel like C would work just fine.

kernel/genarch/src/drivers/pl011/pl011.c
```diff
@@ -59,9 +59,13 @@ static void pl011_uart_putwchar(outdev_t *dev, wchar_t ch)
 {
 	pl011_uart_t *uart = dev->data;
 
-	if (!ascii_check(ch)) {
+	/* If the userspace owns the console, do not output anything. */
+	if (uart->parea.mapped && !console_override)
+		return;
+
+	if (!ascii_check(ch))
```
le-jzr: I quite dislike mismatched branch bracing.
We have quite a few of those in other places, so obviously that's just me nitpicking,
but I wonder if this style is something we generally consider ok, or if it would make sense
to curb it using ccheck?

tools/autotool.py
```diff
@@ -200,6 +200,10 @@ def get_target(config):
 		platform = config['PLATFORM']
 		target = "arm-helenos"
 
+	if (config['PLATFORM'] == "arm64"):
```
le-jzr: Do we really want to (continue to) use this naming convention?
I feel like using arbitrary "our" names instead of conventional ones can only be harmful.

jermar: Yes. arm64 is not as arbitrary as you might think. For example, Linux, FreeBSD and also Windows are using it. It is generally understood.

jermar:
Once again, thanks for this great work!

uspace/srv/hid/input/input.c
setupji: I tried searching for the 'console' boot argument but with no luck. Could you please point me to the relevant code?

kernel/arch/arm64/src/start.S
setupji: A problem here is that most of this function does not run from the address it is linked at, until a switch to the kernel mapping is done by the function. I think the code could be written in C but it then relies on the compiler to never refer to any object using its absolute address. Having this part written directly in assembly errs on the safe side.

kernel/genarch/src/drivers/pl011/pl011.c
setupji: This is not completely clear to me. What is meant by "mismatched branch bracing" here and how should be this snippet ideally formatted?

setupji:
Thanks both for the review. I have posted patches to address the comments (post-merge) in https://github.com/HelenOS/helenos/pull/166.

kernel/arch/arm64/src/start.S
le-jzr: I see.

kernel/genarch/src/drivers/pl011/pl011.c
le-jzr: What I mean is when one branch of the `if` is braced, but the other branch isn't.
For example
```c
if (condition) {
    //... code
} else
   one_statement();
```
or
```c
if (condition)
    one_statement();
else {
    //... code
}
```

In my opinion, either both branches should be braced (even in the single line case) or neither.
But that may be just me.

uspace/srv/hid/input/input.c
jermar: For example, on sun4v, tools\ew.py adds this to the command line:

--prom-env boot-args="console=devices/\\hw\\pci0\\01:01.0\\com1\\a"

One can do [something similar](http://www.helenos.org/wiki/UsersGuide/UserSerialConsole#ia32andamd64) on x86.

On mips32/malta, the boot argument is hardcoded in the kernel as of now.

This boot argument is then copied to sysinfo by the kernel and userspace services such as the input and output services check for it using config_get_value("console").

setupji: Thanks for the pointer, I will look into this.

kernel/genarch/src/drivers/pl011/pl011.c
setupji: I see.

PR #166, opened 2019-04-22
AArch64 post-merge clean up

setupji:
A few clean-up patches to address review comments in https://github.com/HelenOS/helenos/pull/165.

Note: The second patch in the series unifies firmware search in `tools/ew.py` for arm64 and sun4v. It changes the script to first search for firmware binaries by checking an environment variable, `EW_QEMU_EFI_AARCH64` for arm64 and `OPENSPARC_BINARIES` for sun4v. If this fails then the script tries finding the binaries in their expected default locations. If this also fails then an error is reported on stderr and the script exits.

However, I'm thinking this might not be the best approach. Perhaps, it would be better to introduce a similar scheme to how the HelenOS toolchain is searched first in `CROSS_PREFIX` and then `/usr/local/cross/`.

In case of `tools/ew.py`, firmware binaries could be first searched in something like `CROSS_FIRMWARE` + `/usr/local/firmware/` and then default locations for a specific firmware (i.e. Fedora installs AArch64 EDK2 as `/usr/share/edk2/aarch64/QEMU_EFI.fd`).

Script `contrib/qemu/build-from-scratch.sh` could be then updated to download+install arm64 and sun4v firmware in `CROSS_FIRMWARE` (if set) or `/usr/local/firmware/`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,7 +13,6 @@
 /Makefile.common
 /Makefile.config
 /PKG
-/QEMU_EFI_ARM64.fd
 /common.h
 /config.h
 /cscope.out
--- a/boot/arch/arm64/src/main.c
+++ b/boot/arch/arm64/src/main.c
@@ -190,12 +190,10 @@ efi_status_t bootstrap(void *efi_handle_in,
 	 */
 
 	/* Statically check PAGE_SIZE and BOOT_OFFSET. */
-#if PAGE_SIZE != 4096
-#error Unsupported PAGE_SIZE
-#endif
-#if !IS_ALIGNED(BOOT_OFFSET, PAGE_SIZE)
-#error Unsupported BOOT_OFFSET
-#endif
+	_Static_assert(PAGE_SIZE == 4096, "PAGE_SIZE must be equal to 4096");
+	_Static_assert(IS_ALIGNED(BOOT_OFFSET, PAGE_SIZE),
+	    "BOOT_OFFSET must be a multiple of PAGE_SIZE");
+
 	/*
 	 * Dynamically check the memory base. The condition should be always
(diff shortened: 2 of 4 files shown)


le-jzr:
LGTM 👍
As for firmware search, I think the goal should be to make it as hassle-free as possible for the usual cases, which is either distro packaged firmware, or using `build-from-scratch.sh`. What you're suggesting makes sense, though using `CROSS_` in the variable's name doesn't sound right. I associate that with compilers, not with emulators.

jermar:
LGTM. Yes, it would be nice if build-from-scratch.sh attempted to install the EFI firmware into the custom location, as it is now done with the OpenSPARC binaries.

jermar:
Thanks for working on this. I've just given you the commit bit so you can push this yourself :-)

jermar:
Just noticed that the EDK2 firmware might soon end up in the QEMU source tree.

setupji:
> LGTM +1
> As for firmware search, I think the goal should be to make it as hassle-free as possible for the usual cases, which is either distro packaged firmware, or using `build-from-scratch.sh`. What you're suggesting makes sense, though using `CROSS_` in the variable's name doesn't sound right. I associate that with compilers, not with emulators.

> LGTM. Yes, it would be nice if build-from-scratch.sh attempted to install the EFI firmware into the custom location, as it is now done with the OpenSPARC binaries.

Thanks both. I am thinking that the best approach to make this really hassle-free might be making the bootloader to not depend on UEFI/EDK2 as discussed on the mailing list. I will look into it (as time allows).

> Thanks for working on this. I've just given you the commit bit so you can push this yourself :-)

Cheers!

PR #167, opened 2019-05-17
Cleanup lib rtld&elf

matthieuriolo:
- corrects a null pointer failure in dlopen()
- prevents lib/rtld to call exit()
- corrects lib/elf to return errno_t instead of it's own error codes

### DIFF
--- a/uspace/lib/c/generic/dlfcn.c
+++ b/uspace/lib/c/generic/dlfcn.c
@@ -53,6 +53,10 @@ void *dlopen(const char *path, int flag)
 	m = module_find(runtime_env, path);
 	if (m == NULL) {
 		m = module_load(runtime_env, path, mlf_local);
+		if (m == NULL) {
+			return NULL;
+		}
+
 		if (module_load_deps(m, mlf_local) != EOK) {
 			return NULL;
 		}
--- a/uspace/lib/c/generic/dlfcn.c
+++ b/uspace/lib/c/generic/dlfcn.c
@@ -34,6 +34,7 @@
  * @file
  */
 
+#include <errno.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <stddef.h>
@@ -52,7 +53,10 @@ void *dlopen(const char *path, int flag)
 	m = module_find(runtime_env, path);
 	if (m == NULL) {
 		m = module_load(runtime_env, path, mlf_local);
-		module_load_deps(m, mlf_local);
+		if (module_load_deps(m, mlf_local) != EOK) {
+			return NULL;
+		}
+
 		/* Now relocate. */
 		module_process_relocs(m);
 	}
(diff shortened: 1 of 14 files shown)


uspace/lib/c/generic/rtld/module.c
```diff
@@ -293,14 +299,19 @@ void module_load_deps(module_t *m, mlflags_t flags)
 			dm = module_find(m->rtld, dep_name);
 			if (!dm) {
 				dm = module_load(m->rtld, dep_name, flags);
-				module_load_deps(dm, flags);
+				errno_t rc = module_load_deps(dm, flags);
+				if (rc != EOK) {
+					return EOK;
```
jermar: This looks suspicious, should it be return rc?

uspace/lib/c/generic/dlfcn.c
```diff
@@ -53,7 +53,7 @@ void *dlopen(const char *path, int flag)
 	m = module_find(runtime_env, path);
 	if (m == NULL) {
 		m = module_load(runtime_env, path, mlf_local);
-		if (module_load_deps(m, mlf_local) != EOK) {
+		if (m == NULL || module_load_deps(m, mlf_local) != EOK) {
```
jermar: I would split this into two independent if's.

uspace/lib/c/generic/rtld/module.c
matthieuriolo: yes, you're right

uspace/lib/c/generic/dlfcn.c
```diff
@@ -52,7 +53,10 @@ void *dlopen(const char *path, int flag)
 	m = module_find(runtime_env, path);
 	if (m == NULL) {
 		m = module_load(runtime_env, path, mlf_local);
```
jermar: I guess here we don't check for m == NULL again.

uspace/lib/c/generic/rtld/module.c
```diff
@@ -180,13 +180,13 @@ module_t *module_load(rtld_t *rtld, const char *name, mlflags_t flags)
 {
 	elf_finfo_t info;
 	char name_buf[NAME_BUF_SIZE];
-	module_t *m;
+	module_t *m = NULL;
```
jermar: Not really necessary as we unconditionally assign to it a couple of lines below.

uspace/lib/c/generic/rtld/module.c
```diff
@@ -180,13 +180,13 @@ module_t *module_load(rtld_t *rtld, const char *name, mlflags_t flags)
...
 	int rc;
 
 	m = calloc(1, sizeof(module_t));
 	if (m == NULL) {
 		printf("malloc failed\n");
-		exit(1);
+		goto error;
```
jermar: What is the failure mode now? I am thinking whether the exit was not there on purpose because a failure in this code is fatal for the task which is being loaded. So do you get a nicer failure now?

uspace/lib/c/generic/dlfcn.c
jermar: Ah, I should have read this from backwards. Now I see :-)

jermar:
LGTM, but let's wait a couple of days to see if @jxsvoboda has some comments on this

PR #168, opened 2019-05-18
Keyboard layout

matthieuriolo:
This is more of an experimental/fun pull request - This request provides a better implementation for switching between different keyboard layouts.  This request only works properly with the pull request [167](https://github.com/HelenOS/helenos/pull/167)

- Adds a default keyboard layout for uspace. CTRL+F1 still allows to select this layout
- Changes uspace's makefile to create for every kb layout a dynamical library
- Adds a bdsh cmd for listing, getting and setting a keyboard layout (aka help layout list¦get¦set <name>)
- Use a public default service name for srv/hid/input
- Implements IPC request handlers in srv/hid/input for retrieving and setting layout for all keyboards


le-jzr:
Looks fun :) but I don't think defining layouts via code/libraries is practical. So yeah, it's probably better than what we have now, but it doesn't really translate to what we *should* have.

matthieuriolo:
> but I don't think defining layouts via code/libraries is practical

Agreed, but this approach allowed me to extend the existing code instead to rewrite the layouts and api... I assume the "what we should have"-part is something very similiar to linux's keymap?

le-jzr:
I'm not really familiar with those, but probably yes. I'm imagining some kind of plain text data files that get loaded by the keymap server.

le-jzr:
IMO the ideal keymap should be easy to edit, both by hand and by tool, and also should be easy and safe to add from external source (e.g. web download).

le-jzr:
That being said, I'm not against merging this in principle. So long as we are not too attached to it, expecting it to be replaced eventually. I'll review the code later. :)

le-jzr:
Ok, so I've thought about this, and I think we really should not be using solibs to load the keymaps. Pretty much the only reason we would do this instead of having the keymaps hardwired is the ability to add new ones. But given the current stage of HelenOS's development, that's not something we need right now. The use of dynamic libraries also creates extra complication into the build system, which I'm trying hard to make as dead simple as possible.

That said, I like all the other parts of this apart from the dynamic library use. If you change this to only add the management stuff, while keeping the keymaps statically baked into the server, I'd love to merge this. Or you could split out the library portion to keep it as an option later.

matthieuriolo:
> Ok, so I've thought about this, and I think we really should not be using solibs to load the keymaps. Pretty much the only reason we would do this instead of having the keymaps hardwired is the ability to add new ones. 

Which was indeed my original intention :)

> That said, I like all the other parts of this apart from the dynamic library use. If you change this to only add the management stuff, while keeping the keymaps statically baked into the server, I'd love to merge this. Or you could split out the library portion to keep it as an option for later.

Mh, this should be possible. I think we can close this pull request and I will provide a new one in the future

le-jzr:
> Mh, this should be possible. I think we can close this pull request and I will provide a new one in the future

Alright then!

PR #170, opened 2019-06-17
Change type of uspace pointers in kernel from pointer type to numeric

le-jzr:
Unresolved issue: `ccheck` does not allow the `uspace_ptr(type)` construct.

Commit message:
> From kernel's perspective, userspace addresses are not valid pointers,
and can only be used in calls to `copy_to/from_uspace()`.
Therefore, we change the type of those arguments and variables to
`uspace_addr_t` which is an alias for `sysarg_t`.
>
>This allows the compiler to catch accidental direct accesses to
userspace addresses.
>
>Additionally, to avoid losing the type information in code,
a macro `uspace_ptr(type)` is used that translates to `uspace_addr_t`.
It makes no functional difference, but allows keeping the type information
in code in case we implement some sort of static checking for it in the future.

### DIFF
--- a/kernel/generic/include/udebug/udebug_ops.h
+++ b/kernel/generic/include/udebug/udebug_ops.h
@@ -54,7 +54,7 @@ errno_t udebug_args_read(thread_t *t, void **buffer);
 
 errno_t udebug_regs_read(thread_t *t, void **buffer);
 
-errno_t udebug_mem_read(sysarg_t uspace_addr, size_t n, void **buffer);
+errno_t udebug_mem_read(uspace_ptr(void) uspace_addr, size_t n, void **buffer);
 
 #endif
 
--- a/kernel/generic/include/udebug/udebug_ops.h
+++ b/kernel/generic/include/udebug/udebug_ops.h
@@ -54,7 +54,7 @@ errno_t udebug_args_read(thread_t *t, void **buffer);
 
 errno_t udebug_regs_read(thread_t *t, void **buffer);
 
-errno_t udebug_mem_read(uspace_ptr(void) uspace_addr, size_t n, void **buffer);
+errno_t udebug_mem_read(uspace_addr_t uspace_addr, size_t n, void **buffer);
 
 #endif
 
--- a/kernel/generic/src/ipc/irq.c
+++ b/kernel/generic/src/ipc/irq.c
@@ -222,7 +222,7 @@ static void code_free(irq_code_t *code)
  * @return Kernel address of the copied IRQ code.
  *
  */
-static irq_code_t *code_from_uspace(uspace_ptr(irq_code_t) ucode)
+static irq_code_t *code_from_uspace(uspace_ptr_irq_code_t ucode)
 {
 	irq_pio_range_t *ranges = NULL;
 	irq_cmd_t *cmds = NULL;
@@ -322,7 +322,7 @@ static kobject_ops_t irq_kobject_ops = {
  *
  */
 errno_t ipc_irq_subscribe(answerbox_t *box, inr_t inr, sysarg_t imethod,
-    uspace_ptr(irq_code_t) ucode, uspace_ptr(cap_irq_handle_t) uspace_handle)
+    uspace_ptr_irq_code_t ucode, uspace_ptr_cap_irq_handle_t uspace_handle)
 {
 	if ((inr < 0) || (inr > last_inr))
 		return ELIMIT;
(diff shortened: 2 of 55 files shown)


kernel/generic/include/udebug/udebug_ops.h
```diff
@@ -54,7 +54,7 @@ errno_t udebug_args_read(thread_t *t, void **buffer);
 
 errno_t udebug_regs_read(thread_t *t, void **buffer);
 
-errno_t udebug_mem_read(sysarg_t uspace_addr, size_t n, void **buffer);
+errno_t udebug_mem_read(uspace_ptr(void) uspace_addr, size_t n, void **buffer);
```
jermar: How about using uspace_addr_t instead of uspace_ptr(void)?

kernel/generic/src/ipc/irq.c
```diff
@@ -322,7 +322,7 @@ static kobject_ops_t irq_kobject_ops = {
  *
  */
 errno_t ipc_irq_subscribe(answerbox_t *box, inr_t inr, sysarg_t imethod,
-    irq_code_t *ucode, cap_irq_handle_t *uspace_handle)
+    uspace_ptr(irq_code_t) ucode, uspace_ptr(cap_irq_handle_t) uspace_handle)
```
jermar: One of the issues I see is that we are losing a degree of type safety here. Can we do something about that by expanding uspace_ptr into some opaque pointers?

kernel/generic/src/ipc/sysipc.c
```diff
@@ -436,7 +436,7 @@ sys_errno_t sys_ipc_call_async_slow(cap_phone_handle_t handle, ipc_data_t *data,
 		return ENOMEM;
 	}
 
-	errno_t rc = copy_from_uspace(&call->data.args, &data->args,
+	errno_t rc = copy_from_uspace(&call->data.args, data + offsetof(ipc_data_t, args),
```
jermar: Hm, this is also unfortunate that we can't no longer use the compiler to compute the member address.

jermar:
Btw, we could instead do something that Linux does with __user pointer annotations:

  # define __user         __attribute__((noderef, address_space(1)))

We don't use sparse to carry out the checks, but we have ccheck, which could do something similar. At the same time, we wouldn't lose type-safety of the pointers and the compiler will still be able to compute the offsets.

kernel/generic/include/udebug/udebug_ops.h
le-jzr: Sure.

le-jzr:
I'll answer to the three comments above here, since they are all related.

If we expand `uspace_ptr()` to a pointer, we have to ensure that it can't be dereferenced by the kernel. That means either doing the checking ourselves, or the type being opaque. If it's opaque, we can't use compiler to compute the member address anyway, and we have to define that opaque type ahead of time for every argument type. If we go the static checking route, we first need a static type checker for it, and `ccheck` cannot do that by design. Additionally, if we decide to write a static checker for it, we can just as well check the non-pointer declarations instead.

With opaque pointers, we can't check easily whether the right kind of pointer is being fed into `memcpy_to/from_uspace()`, or into anything with `void *` argument for that matter.

If we don't implement an external checker or use predefined opaque pointers, than confusing the pointed-to type is a less severe problem by comparison. The worst case scenario is that the kernel doesn't behave as expected and causes the syscall to fail or the task to crash. It is not a security issue because the kernel is not allowed to trust any pointers or data from uspace anyway. On the other hand, directly accessing an address from uspace is a severe issue because malicious uspace task can provide an invalid address and crash the kernel, or it can provide arbitrary kernel address which kernel then overwrites. This actually happens in `sys_as_area_get_info()`.

There is also the original issue that caused me to work on this, which is that from the perspective of C, pointers and `sysarg_t` are not compatible, and we have to explicitly disable checking of function types to work around that, which is not safe or future-proof as it depends on particular properties of the calling convention. While minor by comparison, pointer arguments would mean we'd have to add a bunch of extra explicit conversions in all syscalls that accept pointers.

With the limitation of C in mind, I don't see a perfect solution.
Though if we switched to using a subset C++, it would be pretty easy.

jermar:
Sounds like we can have at least two of the three: inability to access a userspace pointer and type safety between the userspace pointers. The number of types used by syscalls is imho quite small. On the other hand, this can probably be done per-partes, this being the first part. So basically I am fine with this after uspace_ptr(void) is changed to uspace_addr_t everywhere.

le-jzr:
There's still the issue of `ccheck` to solve. @jxsvoboda ?

le-jzr:
> Sounds like we can have at least two of the three: inability to access a userspace pointer and type safety between the userspace pointers.

I don't think that's quite true. I don't see how you can achieve type safety between userspace pointers without also allowing them to be passed to `void *`.

PR #171, opened 2019-07-03
C++ stdlib: <future>

Dzejrou:
This PR adds the entirety of the <future> header, namely:

* std::future
* std::promise
* std::async
* std::packaged_task
* std::shared_future
* Some auxilliary types and functions
* Tests covering this header

The only thing that isn't currently working (that I am aware of) are ::set_X_at_thread_exit member functions which require thread_atexit-like semantics. Included are also fixes for minor bugs that were found while working on this header.


uspace/lib/cpp/include/__bits/thread/shared_state.hpp
```diff
@@ -0,0 +1,409 @@
...
+            /**
+             * TODO: This member function is supposed to be marked
+             *       as 'const'. In such a case, however, we cannot
+             *       use the underlying fibril API because these
+             *       references get converted to pointers and the API
+             *       does not accept e.g. 'const fibril_condvar_t*'.
+             *
+             *       The same applies to the wait_for and wait_until
+             *       functions.
+             */
```
le-jzr: Why not use `const_cast`? This is what would be called "interior mutability" in Rust parlance.

uspace/lib/cpp/include/__bits/thread/async.hpp
```diff
@@ -0,0 +1,146 @@
...
+            using result_t = result_of_t<decay_t<F>(decay_t<Args>...)>;
+
+            bool async = (static_cast<int>(policy) &
+                          static_cast<int>(launch::async)) != 0;
+            bool deferred = (static_cast<int>(policy) &
+                             static_cast<int>(launch::deferred)) != 0;
+
+            /**
+             * Note: The case when async | deferred is set in policy
+             *       is implementation defined, feel free to change.
```
le-jzr: What do other standard library implementations do? IMO if there is no reason to do otherwise, we should match their behavior for compatibility.

uspace/lib/cpp/include/__bits/thread/async.hpp
```diff
@@ -0,0 +1,146 @@
...
+                */
+                return future<result_t>{
+                    new aux::deferred_shared_state<
+                        result_t, F, Args...
+                    >{forward<F>(f), forward<Args>(args)...}
+                };
+            }
+
+            /**
+             * This is undefined behaviour, let's be nice though ;)
```
le-jzr: If the spec says it's undefined behavior, and we detect it, it should abort the program. Allowing undefined behavior to go undetected "to be nice" is a terrible idea. In the case of C++, throwing `std::invalid_argument` is probably the way to go?

uspace/lib/cpp/include/__bits/thread/packaged_task.hpp
```diff
@@ -0,0 +1,228 @@
...
+#include <utility>
+
+namespace std
+{
+    /**
+     * 30.6.9, class template packaged_task:
+     */
+
+    template<class>
+    class packaged_task; // undefined
```
le-jzr: What does "undefined" mean here?

uspace/lib/cpp/include/__bits/thread/shared_state.hpp
```diff
@@ -0,0 +1,409 @@
...
+     * But since we have no plan (nor need) to make those,
+     * this approach seems to be the best one.
+     *
+     * Also note that unlike the parent class shared_state,
+     * we do not need to specialize these for void. This is because
+     * the difference in class contents are handled by the parent
+     * specialization and setting the value can be handled easily
+     * by if constexpr and checking for the equivalence of the
+     * R template parameter and void.
+     */
```
le-jzr: 👍

uspace/lib/cpp/include/__bits/thread/shared_state.hpp
```diff
@@ -0,0 +1,409 @@
...
+     *      (as those have ::mark_set member functions).
+     *   3) If not done already, register a function
+     *      executing all these in the thread_atexit function
+     *      once that is implemented.
+     */
+
+    template<class R>
+    void set_state_value_at_thread_exit(shared_state<R>* state)
+    {
+        // TODO: implement
```
le-jzr: Probably should abort/throw an exception instead of silently doing nothing.

uspace/lib/cpp/include/__bits/thread/shared_state.hpp
```diff
@@ -0,0 +1,409 @@
...
+    }
+
+    template<class R>
+    void set_state_exception_at_thread_exit(shared_state<R>* state)
+    {
+        // TODO: implement
```
le-jzr: Same as above.

uspace/lib/cpp/src/__bits/runtime.cpp
```diff
@@ -206,4 +206,10 @@ namespace __cxxabiv1
     // Needed on arm.
     extern "C" void __cxa_end_cleanup()
     { /* DUMMY BODY */ }
+
+    extern "C" int __cxa_thread_atexit(void(*)(void*), void*, void*)
+    {
+        // TODO: needed for thread_local variables
```
le-jzr: Probably should abort/throw exception.

uspace/lib/cpp/include/__bits/thread/packaged_task.hpp
Dzejrou: It means that the base template does not exist and only explicitly created specializations can be used. In this case it is used to require the template parameter to be a function/callable, denoted by R(Args...) for a given return type R and argument types Args...

le-jzr: Should probably be explained in the comment. Just the word "undefined" doesn't tell me how or why. :)

uspace/lib/cpp/include/__bits/thread/async.hpp
Dzejrou: I do not know, but I think that does not matter. The outside behaviour of both the deferred and async policies is identical (and we are supposed to choose one here). However, the async policy carries a risk (resource busy - fibril/thread unavailable) while the deferred policy does not (it simply executes the function when the result is retrieved). This IMHO makes the deferred policy superior for a default (no chance of surprises).

uspace/lib/cpp/include/__bits/thread/async.hpp
Dzejrou: > Allowing undefined behavior to go undetected "to be nice" is a terrible idea.

Is it? I am not aware of other implementations failing/crashing/reporting when they encounter undefined behaviour. The libstdc++ implementation on my tumbleweed does not crash, it asynchronously executes the given functor.

Since what is visible from the outside should be and is the same for all policies that I'm aware of, I think that falling back to a default here is a pretty reasonable thing to do (the reason for the deferred policy being the default can be read in the other commend - basically no risk of surprises).

uspace/lib/cpp/include/__bits/thread/async.hpp
le-jzr: I see. That makes sense. Could you add that rationale to the comment?

uspace/lib/cpp/include/__bits/thread/async.hpp
le-jzr: Is it really undefined behavior? In C and C++, "undefined behavior" has very specific meaning. It means that any program that touches it loses all semantic meaning. It can start WW III. It can make demons fly out your nose. It can execute the functor as usual. It can do one at those at random in every run. The point is, undefined behavior is the ultimate bad and if you already detect it, there is no reason ever not to make it crash loudly.

le-jzr: If you think that's overreacting, just remember that GCC will be happy to just arbitrary silently remove your code without telling you, whenever it detects that it would invoke undefined behavior. -_-

Dzejrou: Actually, the definition of undefined behaviour is (1.3.24, defns.undefined):

> behaviour for which this International Standard imposes no requirements

The standard explicitly allows (among others):

* ignoring the situation completely
* behaving during translation or execution in a documented manner (with or without a message)
* terminating translation or execution (with a message)

Any of these are usable, but since - as I previously noted - the outside effect of the policy should not be visible to the user and libstdc++ itself does not fail in this case, I think that we should keep the deferred default here (plus it helps compatibility with other implementations).

uspace/lib/cpp/include/__bits/thread/shared_state.hpp
Dzejrou: Good catch, forgot about that, will add it.

uspace/lib/cpp/include/__bits/thread/async.hpp
Dzejrou: Will do.

uspace/lib/cpp/include/__bits/thread/packaged_task.hpp
Dzejrou: Doesn't it? It does right what it says on the box, that template is not defined ;) But sure, I can add a comment.

Dzejrou: Actually, I think that exact comment is taken straight from the standard (but still, I will change it).

uspace/lib/cpp/include/__bits/thread/async.hpp
le-jzr: "imposes no requirements" means what I said. It can have arbitrary nonlocal effects. Compiler optimizers heavily utilize any such lack of requirements by assuming "undefined behavior can't happen". Any program that involves undefined behavior in any way is implicitly buggy in unforeseeable ways and should not be accepted. Compatibility doesn't play a role because where undefined behavior is concerned, the program doesn't have any valid meaning.

Dzejrou: Well, consider the following scenario:

Implementation X adds std::launch::thread_pool, which when used with std::async causes std::async to spawn the asynchronous task in a predefined thread pool as an optimization. The behaviour will be the same (sans the optimization) for us given the semantics of std::async and it is actually _defined behaviour_ for that implementation - this is because the standard actually says that undefined behaviour here is triggered if the launch policy is not defined by the standard _and_ not defined by the implementation (so it can be undefined behaviour on some implementation while defined behaviour on others).

If we leave it as it is, program using std::async::thread_pool written for implementation X and ported over will still work as it did before (sans the optimization of course).

Dzejrou:
> Do we really want to just comment out the test? IMO failing test is what we actually want for something that doesn't work yet.

Well, when invoked it will now abort.

uspace/lib/cpp/src/__bits/test/future.cpp
```diff
@@ -0,0 +1,351 @@
...
+        test("default constructed promise has state", p1.__state());
+
+        std::promise<int> p2{};
+        auto* s1 = p1.__state();
+        auto* s2 = p2.__state();
+        p2.swap(p1);
+        std::swap(s1, s2);
+
+        test_eq("swap switches states pt1", s1, p1.__state());
+        test_eq("swap switches states pt2", s2, p2.__state());
```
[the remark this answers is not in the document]
Dzejrou: Actually, it means part 1 and part 2 (i.e. two assertions that are part of one test).

uspace/lib/cpp/include/__bits/memory/shared_payload.hpp
```diff
@@ -90,43 +77,42 @@ namespace std::aux
...
-                  deleter_{deleter}, refcount_{1}, weak_refcount_{1}
+                  deleter_{deleter}
             {
                 alloc.construct(data_, forward<Args>(args)...);
             }
 
             void destroy() override
             {
-                if (refs() == 0)
+                if (this->refs() == 0)
```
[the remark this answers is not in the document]
Dzejrou: Exactly, had to be added in several places as the template shared_payload now inherits from a non-templated type refcount_obj (because I wanted to move that logic out of smart pointers and reuse it in futures and promises).

uspace/lib/cpp/include/__bits/thread/future_common.hpp
```diff
@@ -0,0 +1,123 @@
...
+         * the inner stored type and the return type of
+         * the get() member function, respectively.
+         */
+
+        template<class T>
+        struct future_inner: aux::type_is<T>
+        { /* DUMMY BODY */ };
+
+        template<class T>
+        struct future_inner<T&>: aux::type_is<T*>
```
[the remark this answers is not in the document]
Dzejrou: Basically, std::future<R&> contains a reference to some object, but this reference is added _after_ that future is created (i.e. it cannot be actual reference, because that has to be initialized immediately). This is a trick I used to avoid code duplication (combined with constexpr if which adds dereference of the pointer, see std::future<R>::get()) so that there is one implementation for R, R& and void even though their workings are different.

If your question is to how the metafunction actually works, we basically tell the compiler two things:

* for any type T, return the type T
* for a type that is a reference to some type T, return pointer to that type T

Given the rules of the language, the compiler is obligated to check the more specialized template first and if it fits, returns the pointer type, otherwise falls back to the base template and acts as identity. As an example, the following static asserts pass:

```cpp
static_assert(std::is_same_v<int, std::aux::future_inner_t<int>>);    // identity
static_assert(std::is_same_v<int*, std::aux::future_inner_t<int&>>);  // specialized version
static_assert(std::is_same_v<int*, std::aux::future_inner_t<int*>>);  // identity again
```

uspace/lib/cpp/include/__bits/thread/shared_state.hpp
```diff
@@ -0,0 +1,429 @@
...
+            }
+
+        protected:
+            future_status timed_wait_(aux::time_unit_t time) const override
+            {
+                /**
+                 * Note: Currently we have no timed join, but this
+                 *       behaviour should be compliant.
+                 */
+                aux::threading::time::sleep(time);
```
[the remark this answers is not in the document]
Dzejrou: Then we're wasting time. When I was writing this we had no functionality like that for fibrils so I went with a crude yet standard compliant approach, but I agree this should be improved upon.

uspace/lib/cpp/include/__bits/thread/future_common.hpp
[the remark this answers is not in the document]
jermar: Thanks for the explanation.

uspace/lib/cpp/src/__bits/test/future.cpp
```diff
@@ -0,0 +1,351 @@
...
+        std::promise<int> p3{std::move(p1)};
+        test_eq("move construction state moved", s1, p3.__state());
+        test_eq("move construction source empty", p1.__state(), nullptr);
+
+        p1 = std::move(p3);
+        test_eq("move assignment state move", s1, p1.__state());
+        test_eq("move assignment source empty", p3.__state(), nullptr);
+
+        p1.set_value(42);
+        test("set_value marks state as ready", s1->is_set());
```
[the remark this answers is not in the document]
Dzejrou: Now that's a tricky one. I can add a check for that if you wish, but the reason for not checking it was that we don't have exception support yet. This is because that if one calls `set_value` when the shared state is already marked as ready, an exception with the error code `std::future_errc::promise_already_satisfied` is thrown, but right now throwing an exception results to `abort` being called (and a message being printed to stderr). Because of the abort it was not feasible to test for it, but if you want I can add a check for `!is_set()` on the shared state before `set_value` is called (though it is IMHO redundant because of the exception/abort).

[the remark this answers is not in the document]
jermar: I think there are more places where you check the final value, but don't check the initial value, and so you don't actually know whether calling a function results in the desired change. From your reply I understand that calling is_set() before set_value() is okay provided the tested code is correct? Then I'd do it. Unfortunately, I didn't track all the other occurrences that follow the same pattern.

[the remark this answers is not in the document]
Dzejrou: > From your reply I understand that calling is_set() before set_value() is okay provided the tested code is correct?

The `is_set` member function is purely internal because it is implemented on the shared state (which itself is an implementation detail), so there should be no problem testing it before. I will check when I'm home just to be sure, though I might just merge this and add this kind of testing to my TODO list in my cpp-qol [0] branch which I'm working on.

[0] https://github.com/Dzejrou/helenos/tree/cpp-qol

PR #172, opened 2019-07-18
Shutdown

matthieuriolo:
This PR provides the missing mechanism for calling halt/reboot in uspace. It does NOT tackle the issues stated in [525](http://www.helenos.org/ticket/525) or  [414](http://www.helenos.org/ticket/414) !

- Some restructering (file names, contents and locations appeared to be inconsistent)
- Creating a systemcall for shutdown
- Adding application for shutdown

I haven't seen any better way to access the kernel task as with `task_find_by_id(1);`. If anyone has got an idea how to obtain the kernel task or the ID in a better way, please let me know.


kernel/generic/src/main/shutdown.c
```diff
@@ -52,5 +97,92 @@ void reboot(void)
...
+	if (mode != SHUTDOWN_HALT && mode != SHUTDOWN_REBOOT) {
+		return EOK;
+	}
+
+	sys_shutdown_arg_t *arg = malloc(sizeof(sys_shutdown_arg_t));
+	if (arg == NULL) {
+		return ENOMEM;
+	}
+
+	//TODO: find a better way for accessing the kernel task
```
jermar: There is currently no other way probably, simply because there was no need for it. Moreover, global task IDs will be phased out at some point in favour of local IDs (capability handles), so this will cease to work. Not tomorrow though, but you've been warned :-)

matthieuriolo:
I see, I was not aware of this and thought this could be handy. From what I can tell the branch is only available on [launchpad](https://code.launchpad.net/~werkov/helenos/system-daemon) ? And the last time it was updated was in 2016? Is someone still working on it and is it known as to why it wasnt merged?

matthieuriolo:
I've seen in the meanwhile that the thesis contains a small list of todos (page 47 - Chapter 5. Conclusion). But none of them speaks of the arch problems which you mentioned. I will have a look at the branch but honestly I'm not sure if I can make it work. I assume you used [git-remote-bzr](https://github.com/felipec/git-remote-bzr) for importing bazaar to git?

Considering this PR, it's up to you to decide to keep it or to reject it. But I believe you already made up your mind on that topic ^^

jermar:
I think the conclusion mentions it in this part:

> This is more configurable than the original hardcoded version (although it brings the burden of keeping configurations for mutliple platforms, which was not addressed in the thesis).

On top of that, it might require some non-trivial amount of effort to reconcile the thesis with the current master.

As for this PR, I am afraid it cannot be merged. It basically gives an equivalent of a power switch to every task, which is quite dangerous and not what you want when you type 'shutdown' or 'reboot'.

matthieuriolo:
> I think the conclusion mentions it in this part:
> 
> > This is more configurable than the original hardcoded version (although it brings the burden of keeping configurations for mutliple platforms, which was not addressed in the thesis).

Ah yes, what you said stroke me way more serious than that :)

> As for this PR, I am afraid it cannot be merged. It basically gives an equivalent of a power switch to every task, which is quite dangerous and not what you want when you type 'shutdown' or 'reboot'.

I can see that and it's understandable to why. I just saw that there is no way to do this in uspace but there was a way in kernel space. That's why I tried to create the bridge. But you're right, it is a footgun

PR #173, opened 2019-07-22
Kb layout switch

matthieuriolo:
As promised to @le-jzr in PR [168](https://github.com/HelenOS/helenos/pull/168) here is a new PR which keeps the hardcoded layouts but provides the needed IPC and command for switching between the layouts. LCTRL-F1 still allows you to switch to the US layout. Unlike to the previous PR the default keyboard can not any longer be configured with `make config`. The idea is that in the future (once keymaps are implemented) the default layout/map will still be hardcoded to ensure that there is at least one keymap around.

The following wikipages need to be updated:
http://www.helenos.org/wiki/UsersGuide/Console
http://www.helenos.org/wiki/Tutorial

Sidenote, both wiki pages are missing the layout "ar" for arabic and it only works if the left CTRL is pressed

-----------------------------

For the keymap, I looked around and found out that it appears every operating system seems to be doing their own thing. I looked at FreeBSD, Arch Linux, Mac OS X and Windows 10. From what I can tell, the linux keymap is the most powerful while FreeBSD has the most simplest one. Linux provides the functionality:
 - Including/Loading other keymaps
 - Single character mapping
 - Auto uppercasing (ex.: +a will automatically map to A when shift/caps is pressed)
 - Mapping to a string instead of a single character
 - Strings can contain a newline and will therefore execute in a shell
 - Dead keys
 - able to differentiate between left/right ctrl or shift

Looking at those functionality I dont believe the structure of HelenOS is able to cope with them. And also probably shouldnt provide those functionality. As result of that I would have suggested to implement a slighty changed version of the FreeBSD keymap

```
# scancode base shift ctrl ctrl+shift alt alt+shift alt+ctrl alt+ctrl+shift   lock state
0x1e 'a' 'A' 1 1 'a' 'A' 1 1 caps
96 gravel gravel - - - - - - -
gravel          32   '`'      'a'  'à'      0xE8   'è'      'i'  0xEC
```
https://github.com/freebsd/freebsd/blob/master/share/vt/keymaps/fr.acc.kbd

 - control characters have to be expressed by their numeric value
 - lock state can be: -, caps, num, both (instead of C, N, B, O)
 - custom "accent names" (must start with a letter; single world)
 - accent mapping is done by a key value array (instead of using brackets)
 - the minus symbol can be used to dismiss an option (instead of nop)


uspace/srv/hid/input/input.c
```diff
@@ -55,6 +55,7 @@
 #include <stdlib.h>
 #include <str.h>
 #include <str_error.h>
+#include <dlfcn.h>
```
jxsvoboda: You probably forgot to remove this.

uspace/app/layout/layout.c
```diff
@@ -0,0 +1,178 @@
...
+	    "\t%s list\tlists all layouts\n"
+	    "\t%s get\t displays currently set layout\n"
+	    "\t%s set <layout>\tchanges to the new layout\n",
+	    cmdname, cmdname, cmdname, cmdname);
+}
+
+/* lists all available kb layouts */
+static errno_t list_layout(void)
+{
+	/* TODO: replace this with the content of the folder containing the keymaps */
```
jxsvoboda: I think rather the TODO should say we should query the server for available maps.

uspace/srv/hid/input/input.c
```diff
@@ -205,33 +264,9 @@ void kbd_push_event(kbd_dev_t *kdev, int type, unsigned int key)
...
-		layout_destroy(kdev->active_layout);
-		kdev->active_layout = layout_create(layout[2]);
-		return;
-	}
-
-	if ((type == KEY_PRESS) && (kdev->mods & KM_LCTRL) &&
-	    (key == KC_F4)) {
-		layout_destroy(kdev->active_layout);
-		kdev->active_layout = layout_create(layout[3]);
+	if ((type == KEY_PRESS) && (kdev->mods & KM_LCTRL) && (key == KC_F1)) {
```
jxsvoboda: Please don't take away the ability to switch between layouts via keyboard shortcuts.

matthieuriolo:
> while it's nice to have the ability to query and set the layout via the command line in addition to keyboard shortcut, it's not a replacement, so that functionality should still be there, please don't remove it. We could do a shortcut for cycling the layouts.

I would strongly advice against having the functionality of "cycling". This would not work well in the future when the keymaps are implemented as files. I doubt that anyone really needs this feature? The old implementation has the obvious problem of a limited support of 9 keymaps. I think it would be more important to free up those CTRL-combinations for more important functionality. Ex.: Shutdown, reboot, new BDSH Terminal, etc.

> Even then we could keep the Fns for some select keymaps (which could be made configurable in the future)

The prober way to approach this would be to implement shortcut-functionality. Selecting a specific keymap would be a shortcut. However this is heavily depended on what type of keymap we use

They way I see it, this is legacy code which should be removed (better sooner as later)

uspace/app/layout/layout.c
matthieuriolo: When I implemented this function I thought about this and came to the conclusion that querying the server is unnecessary complicated with no benefit. It's an extra IPC call with no real advantage? I think it's easier if the server receives a path with the location of the keymap. Displaying a list (and translating that list) is a job done by the app

jxsvoboda:
> I would strongly advice against having the functionality of "cycling". This would not work well in the future when the keymaps are implemented as files. I doubt that anyone really needs this feature?
What I meant is cycling between a user-preselected set of keymaps, not between all available keymaps. This feature is present just about in every desktop environment and most people who use multiple keymaps use it.

> The old implementation has the obvious problem of a limited support of 9 keymaps. I think it would be more important to free up those CTRL-combinations for more important functionality. Ex.: Shutdown, reboot, new BDSH Terminal, etc.
Well we don't have anywhere near 9 keymaps right now.

> The way I see it, this is legacy code which should be removed (better sooner as later)
Code that provides useful functionality should not be removed without providing appropriate replacement.

matthieuriolo:
I believe this merge request has been forgotten. Keep in mind that french keyboard has been added in the meanwhile.

le-jzr:
Sorry about that, I've been super busy in the past forever and completely forgot about this. Also, there seem to be conflicts now, which need to be resolved. One is caused by the switch to Meson build system, should be an easy fix but let me know if you need hints. Other than that, @jxsvoboda is currently working on the general UI area, so I defer to him.

matthieuriolo:
> Sorry about that, I've been super busy in the past forever and completely forgot about this

Nothing to be sorry about. I do know and understand that you guys are powerhorsing an operating system in your free time and therefore such small changes get forgotten 😃

> Also, there seem to be conflicts now, which need to be resolved. One is caused by the switch to Meson build system, should be an easy fix but let me know if you need hints

Just that we are on the same page: You want me to merge master, resolve the issues and push force the new pull request? I remember that jakub once told me that I should prevent unnecessary merge commits and that it's better to let you guys handle such small conflicts. Just asking

jermar:
I am not sure I said exactly that :-) IMHO the best approach would be rebasing your change on top of the current master and force-pushing it into your branch. This way there won't be the extra commit which merges master into your branch. So in total two merge commits will be avoided.

uspace/app/layout/layout.c
```diff
@@ -0,0 +1,158 @@
...
+
+	printf(
+	    "Usage: %s\n"
+	    "\t%s list             lists all layouts\n"
+	    "\t%s get              displays currently set layout\n"
+	    "\t%s set <layout>     changes to the new layout\n",
+	    cmdname, cmdname, cmdname, cmdname);
+}
+
+static async_sess_t *hid_exchange_start()
```
jxsvoboda, line 61: Why don't you use input_open() to get an input_t object and use that to talk to the input server?

jxsvoboda: Also, print_help() is not valid syntax in ANSI/ISO C.

uspace/lib/c/generic/io/input.c
```diff
@@ -201,5 +202,66 @@ static void input_cb_conn(ipc_call_t *icall, void *arg)
...
 
+/**
+ * Retrieves the active keyboard layout
+ * @param sess Active session to the input server
+ * @param layout The name of the currently active layout,
+ *        needs to be freed by the caller
+ * @return EOK if sucessful or the corresponding error code.
+ *         If a failure occurs the param layout is already freed
+ */
+errno_t input_layout_get(async_sess_t *sess, char **layout)
```
jxsvoboda, line 213: Why does this not have input_t * as the first argument?

uspace/lib/c/generic/io/input.c
```diff
@@ -201,5 +202,66 @@ static void input_cb_conn(ipc_call_t *icall, void *arg)
...
+	return rc;
+}
+
+/**
+ * Changes the keyboard layout
+ * @param sess Active session to the input server
+ * @param layout The name of the layout which should be activated
+ * @return EOK if sucessful or the corresponding error code.
+ */
+errno_t input_layout_set(async_sess_t *sess, const char *layout)
```
jxsvoboda, line 250: Why does this not have input_t * as the first argument?

uspace/lib/c/include/ipc/input.h
```diff
@@ -38,7 +38,9 @@
 #include <ipc/common.h>
 
 typedef enum {
-	INPUT_ACTIVATE = IPC_FIRST_USER_METHOD
+	INPUT_ACTIVATE = IPC_FIRST_USER_METHOD,
+	INPUT_CHANGE_LAYOUT,
```
jxsvoboda, line 42: For symmetry with GET_LAYOUT this should be named SET_LAYOUT.

uspace/srv/hid/input/input.c
```diff
@@ -107,6 +109,61 @@ static list_t serial_devs;
...
+			layout_change(layout_active);
+			return EOK;
+		}
+	}
+
+	return ENOTSUP;
+}
+
+/* Handler for IPC call INPUT_CHANGE_LAYOUT */
+static void client_change_layout_handler(ipc_call_t *call)
```
jxsvoboda, line 140: Typically the suffix used for these functions would be _srv().

uspace/srv/hid/input/input.c
```diff
@@ -107,6 +109,61 @@ static list_t serial_devs;
...
+
+/* Handler for IPC call INPUT_GET_LAYOUT */
+static void client_get_layout_handler(ipc_call_t *call)
+{
+	const char *layout_name = layout_active->name;
+	size_t length = str_size(layout_name) + 1;
+	ipc_call_t id;
+
+	async_answer_1(call, EOK, length);
+	if (async_data_read_receive(&id, NULL)) {
```
jxsvoboda, line 162: This is wrong. You can't reply to *call with EOK first. You first need to call async_data_read_receive() and if it fails, you need to reply with error to call.

.gitignore
```diff
@@ -5,4 +5,4 @@
 /tools/*.pyc
 tags
 tools/sycek/
-tools/xcw/demo/viewer
+tools/xcw/demo/viewer
```
jxsvoboda, line 8: Why did you remove the trailing newline?

uspace/app/layout/layout.c
matthieuriolo: > Why don't you use input_open() to get an input_t object and use that to talk to the input server?

I remembering looking into that function and coming to the conclusion that this is an overkill (considering the refactoring it needs). The function is needed for a long time session while I need a single call.


> print_help() is not valid syntax in ANSI/ISO C

I'm sorry I cannot follow. What's wrong with the naming?

.gitignore
matthieuriolo: that's probably an accident from switching from the old make build to meson. My original code was written for make and I had to exclude the layout folder

uspace/app/layout/layout.c
le-jzr: > > print_help() is not valid syntax in ANSI/ISO C
> 
> I'm sorry I cannot follow. What's wrong with the naming?

I think what he means is that a function with no arguments is declared as `print_help(void)`. Empty parentheses mean undeclared list of arguments (or declared elsewhere), and is not accepted in modern standard C, although GNU still accepts the syntax as an extension.

matthieuriolo:
@jxsvoboda I made the necessary changes except for
1. "Why does this not have input_t * as the first argument?"
2. "I think rather the TODO should say we should query the server for available maps"

matthieuriolo:
> My suggestion is that, as future work, the code to look up and connect the input service should be moved to input_open() 

Ah right, so the issue is the code duplication of creating an exchange session for the input service? Because when you look at the structure of input_t and what input_open() does, they serve a completely different purpose - it's a wrapper, not just a connection. This would add complexity (passing useless NULL values to input_open()) for lil reduction of code duplication. Not sure if it is worth

jxsvoboda:
It's not just a case of code duplication, but also of proper abstraction/encapsulation. Each service (connection to a service) should be represented by a class (translated into C, i.e. structure + functions/methods). This means the users of the API need not include <async.h> and need not know there is any IPC going on, they should be (almost) completely shielded from this. I know might not be  the case with all current code, but that's life: best practices improve and it takes time for the code to catch up. Nevertheless, when writing new code, we should observe current best practices. Such a class would typically have an interface something like:
  * method/function myservice_open(const char *svcname, myservice_t **)
  * errno_t myservice_dowhatever(myservice_t *, ...)
  * void myservice_close(myservice_t *)
I agree that these are two completely different use cases. We could make these into two separate interfaces. I.e. they could be abstracted by a different class and also separated on sever side - into two separate services.

le-jzr (approved):
I apologize if this was waiting for my review. It's been a crazy year.
Looks good. Just needs to resolve the new conflicts.

matthieuriolo:
@le-jzr I rebased the branch and fixed the conflicts

matthieuriolo:
@le-jzr @jermar @jxsvoboda guys, it's getting ridiculous. I'm tying me best to have patient and to be understanding of your situations. But it's now 2 Years... You need to find someone who takes care of the opensource part of Helenos. Obviously you are not fit for this task...

jxsvoboda:
@matthieuriolo was that a necessary comment? A simple reminder would have sufficed. Also, if you want to get attention, a message to the mailing list is a good idea. Also a general note on contributing: If you have an idea that you'd like to implement, the *first* step that I would suggest is to discuss it on the ML so that we can first flesh out the design. It's better than shooting a pull request and than having people to deal with a change that might have fundamental design issues.

Sorry for not tending to the PRs, but one tends to avoid uncomfortable stuff (which harks back to the above).

Now back to the matter. I was almost decided to merge this and clean up the remaining issues, but when I tested it, I found out that when you change the layout using keyboard shortcuts, output of 'layout get' does not change. Then I realized the problem. Keyboard layout is currently a property of a keyboard device. You introduced layout_active, a global variable. When layout is set using 'layout' command, layouts of all keyboard devices are changed. But layouts of individual keyboard devices can be changed using keyboard shortcuts and this is not reflected in output of 'layout get'.

I consider the ability to have different layout for different keyboard device a very useful feature. Imagine having two keyboards plugged in, each with different key labels and the matching layout set in the OS.

So if the command-line layout switching is to work correctly, it should be able to list and set the layout of each individual keyboard device.

PR #177, opened 2020-03-06
System daemon

matthieuriolo:
I worked on Michal Koutný [thesis](http://www.helenos.org/doc/theses/mk-thesis.pdf) for implementing a proper bootstrap system. The merge request is therefore largely based on his work which I got from [launchpad](https://code.launchpad.net/~werkov/helenos/system-daemon).

First of all, _the branch isnt stable yet and not ready to merge_. But I think I've progressed enough that the code should be reviewed. Preferrable a new branch should be created in the official helenos repo - this would allow other people to tinker around with it.

# What I did
Essentially I spent an awful long time on changing the old code to be capable of working with the new API. This included serveral bugfixes. The only real game changer which I added is the `ConditionArchitecture` flag. Originally I had the idea of building the units files with python but decided against it as it would mean lil effect but a lot of work. This newly introduced flag prevents the services s3c24xx_uart and s3c24xx_ts from being started when sysman executes the units on a non-arm32 machine. This is probably already enough to make sysman ready for all the supported  architectures.

# During the review special attention should be given to

[2f04bdd7](https://github.com/matthieuriolo/helenos/blob/system_daemon/uspace/lib/c/generic/libc.c#L71-L83) - the issue was that taskman tried to establish a connection to itself. As result of that the fibril from taskman got deadlocked and this prevented all others application to be started. I didnt find a better way of detecting the taskman process then by calling the `stats_get_task()` function. It strikes me as ugly but it only occurs for the application from INIT_TASKS.

[63a3276](https://github.com/matthieuriolo/helenos/blob/system_daemon/uspace/srv/locsrv/locsrv.c#L353-L356) - The devman unit did not start automatically as this piece of code promises it. I had to explicitly add devman. Since I'm not informed enough how the driver architecture changed I would prefer not to delete this - even though I believe it to be unnecessary and deadcode.

[0a8f070](https://github.com/matthieuriolo/helenos/blob/system_daemon/uspace/app/trace/ipcp.c#L304-L309) - The initial phones has been changed from NS to taskman. I'm not sure if the application trace still is doing what it is supposed to do - I never tested it.

[e0e7eba](https://github.com/matthieuriolo/helenos/commit/e0e7ebacb0769fd1a0bc760347da23ebb0b909ed#diff-5a6de06e1a3a58dd649f7b048cd8e50cL300-L323) - The function `ns_task_disconnect()` has been deleted. I'm not sure what the implication of that is.

# Known Issues

Including the xhci driver will cause `sysctl shutdown` to fail with an infinite amount of "cpu0: unhandled IRQ 11" errors. 

Currently /w/data and /w/cfg are not mounted. I intend on working on the implementation of the `OnFailure` flag - allowing sysman to mount those disk or fallback on the live image mode.

Sysman does not detect proberly if a task fails or if it was stopped by the user. As example, if you start the system and close the barber application (X-button of the window), sysman will mark this unit as failed. This causes problem when trying to execute the shutdown process as sysman will have a failed assertion. This issue is closely related to the next point.

Sysman does not provide any possibility to restart a failed unit - hence missing the `Restart=Always` option. It's such an important feature that I would call it a bug when not included.

Sysman does not gracefully halt the tasks. The original code of Michal contained [a function](https://bazaar.launchpad.net/~werkov/helenos/system-daemon/view/head:/uspace/lib/c/generic/async.c#L826) which would allow the tasks to be properly halted but I was not capable of importing it into the new code base. Unfortunately/luckely this function never was used in the original code. Mounts will be properly unmounted though! I conducted some experiments of finding a way for stopping tasks gracefully but ended up with an unsatisfied solution. Essentially I had a solution in which taskman had two sessions to every task. The initial session and the session to stop the task.

# Other important mentionings
The branch has been forked before the change to the meson build system. I made sure not to import the "/root" mount. When I started the fork it looked like something which made no sense and just overcomplicates the problem :) My notes&todos can be found on [gist](https://gist.github.com/matthieuriolo/9d4f05ed8d52615f2b430b1c9b77af84). This includes the revision numbers of the commit which I cherrypicked from Michal


le-jzr:
Nice to see someone is trying to revive this. :)

le-jzr:
@matthieuriolo Seeing as this is a huge change, do you think there are any parts that naturally fall out as independent and that could be merged separately?

matthieuriolo:
> @matthieuriolo Seeing as this is a huge change, do you think there are any parts that naturally fall out as independent and that could be merged separately?

One thing which comes to my mind is the the separation of [taskman and ns service](https://github.com/HelenOS/helenos/pull/177/commits/0a8f070c890fb7e10df2f1aa71289773d26926dc). I dont think it will as easy as to cherrypick this commit since I bugfixed some stuff afterwards. I tried to make every commit as clean as possible but... well ... I was out of my league and tried my best :)

The advantage of this separation is the possibility to "subscribe to task events". Essentially other tasks can receive notification if a task stops/aborts


Besides that there are two new libraries. [dynamic arrays](https://github.com/matthieuriolo/helenos/blob/system_daemon/uspace/lib/c/generic/adt/array.c) and a [ini parser](https://github.com/matthieuriolo/helenos/tree/system_daemon/uspace/lib/conf). Those probably can be cherrypicked easily.

PR #178, opened 2020-03-09
Fix usbinfo utility usage message typo.

manux81:
This patch fix a typo into usbinfo usage message:
"-r  --hid-report-usages" must be replaced to "-R --hid-report-usages"


vhotspur:
Good catch, thanks!

PR #179, opened 2020-03-10
Add cppcheck extra target to allow static analysis.

manux81:
Cppcheck is a static analysis tool for C/C++ code. It provides unique code analysis to detect bugs and focuses on detecting undefined behaviour and dangerous coding constructs. The goal is to detect only real errors in the code (i.e. have very few false positives). 
The original name of this program was "C++check", but it was later changed to "Cppcheck".
Despite the name, Cppcheck is designed for both C and C++.

To run target just type:
ninja cppcheck > /dev/null
Note: There is a problem with memory.cpp so cppcheck analysis ends about 30% completed.

[helenos_cppcheck_without_memory.cpp.txt](https://github.com/HelenOS/helenos/files/4312297/helenos_cppcheck_without_memory.cpp.txt)

### DIFF
--- a/meson/part/extra_targets/meson.build
+++ b/meson/part/extra_targets/meson.build
@@ -81,6 +81,12 @@ if doxygen.found()
 	])
 endif
 
+if cppcheck.found()
+	run_target('cppcheck', command: ['cppcheck', '--project=' +
+		join_paths(meson.build_root(), 'compile_commands.json'),
+		' -q --force --file-list=*.c* --platform=unix32 --enable=all' ])
+endif
+
 # TODO text-xcw
 #ifeq ($(CONFIG_DEVEL_FILES),y)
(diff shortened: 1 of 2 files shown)


le-jzr:
Awesome!

To be clear, `memory.cpp` is problem in HelenOS or in cppcheck?

manux81:
> Awesome!
> 
> To be clear, `memory.cpp` is problem in HelenOS or in cppcheck?

I just finished to check with last cppcheck built from scratch and now it works!
I think is a cppcheck problem.

le-jzr:
Thanks!

Dzejrou:
Just a note: memory.cpp is the test file for <memory> and as such plays a lot with artificial scopes (to limit the lifetimes of the pointers so that I can check whether stuff gets destroyed properly). Because of that I wouldn't be surprised if the check had a bit of a hard time munching the sources.

See e.g. https://github.com/HelenOS/helenos/blob/master/uspace/lib/cpp/src/__bits/test/memory.cpp#L157

PR #180, opened 2020-03-11
Avoid array 'tmap->tasks[32]' accessed at index 32, which is out of bounds.

manux81:
Found by cppcheck tool

### DIFF
--- a/boot/generic/src/payload.c
+++ b/boot/generic/src/payload.c
@@ -242,7 +242,7 @@ void extract_payload(taskmap_t *tmap, uint8_t *kernel_dest, uint8_t *mem_end,
 
 	tmap->cnt = 0;
 
-	for (int i = 0; i <= TASKMAP_MAX_RECORDS; i++) {
+	for (int i = 0; i < TASKMAP_MAX_RECORDS; i++) {
 		/*
 		 * `task` holds the location and size of the previous component.
 		 */


jermar:
On line 260 there is a check, which will not allow index 32 to be accessed. The construction is a little weird but allows to print an error message if there are too many tasks and halt.

Dzejrou:
> On line 260 there is a check, which will not allow index 32 to be accessed. The construction is a little weird but allows to print an error message if there are too many tasks and halt.

This means the check is performed on every iteration, does it not? What about if the control variable was declared before the for loop that uses it and the check is moved after the loop ends? This way we do not check it on each iteration, we do not risk anything due to the lax for loop condition (I'm not saying there are hidden bugs, but future modifications of the loop can easily introduce them because of this) and we conform with the check. All that at no real price, what do you think? (The for loop is at the end of the scope, so there should not be problems with the control variable having its scope extended, but even so an artificial scope would help there.)

le-jzr:
@Dzejrou That would mean duplicating most of the loop, and comparing two integers is insignificant relative to the other stuff.

@manux81 As @jermar said, the original code is correct, if a bit confusing. The condition of for loop is redundant. Isn't there any way to tell cppcheck that `halt()` never returns? It's already marked such by a GNU attribute, so I'm confused as to why it doesn't understand that.

manux81:
Flow control check needs a loop break after halt function is called to disable error (I already tried and it works) but we can considerate as false positive.

le-jzr:
If we do that, some other tool might report the break as dead code. :) I think we can close this for now. Maybe later versions of cppcheck will add proper handling of noreturn attributes.

PR #183, opened 2020-03-20
Fix memory leaks in nic_wol_virtue_add_impl procedure by free virtue variable.

manux81:


le-jzr:
Don't we also need to free `virtue->data`?

manux81:
Yes, you are right! but when length is equal to zero I can't free virtual->data field. What do you think to move down "if (length != 0)" block?

le-jzr:
> but when length is equal to zero I can't free virtual->data field.

That is not actually true. In that case the field is `NULL`, and `free(NULL)` is a no-op.

manux81:
Ok

manux81:
Fixed

le-jzr:
Thanks!

PR #196, opened 2020-05-07
Avoid use uninitialized variable in ethip_send function.

manux81:
In the ethip_send function, the stacked variable called "frame" is passed to arp_translate not initialized.

### DIFF
--- a/uspace/srv/net/ethip/ethip.c
+++ b/uspace/srv/net/ethip/ethip.c
@@ -174,7 +174,7 @@ static errno_t ethip_send(iplink_srv_t *srv, iplink_sdu_t *sdu)
 	log_msg(LOG_DEFAULT, LVL_DEBUG, "ethip_send()");
 
 	ethip_nic_t *nic = (ethip_nic_t *) srv->arg;
-	eth_frame_t frame;
+	eth_frame_t frame = { 0 };
 
 	errno_t rc = arp_translate(nic, sdu->src, sdu->dest, frame.dest);
 	if (rc != EOK) {


le-jzr:
There's more going on here than meets the eye. As far as I can tell, the respective fields are only used in an output position to initialize them. This is not obvious at first because the `addr*_t` types are defined as array types, which makes them pass-by-reference without the `&` operator. That seems fairly confusing to me, but not necessarily incorrect.

@jxsvoboda Any comments?

jxsvoboda:
I agree completely with le-jzr. We only write to the fields of frame so it should not be necessary to initialize them (except to please some checker that is unaware). The confusing part is that addr48_t is defined as an array. You could try if, alternatively, adding an & when passing the arguments pleases the checker. That would make things more obvious and should not change the semantics (&array == array).

I think we should define these types as struct { array }, too. Nor do I like the name addr48_t, because it's supposed to mean "any 48-bit address", but I'd rather have something like mac_addr_t or eth_addr_t.

manux81:
Yes, yours suggests passed check!

le-jzr:
@manux81 I missed the latter part of your message, probably because you added it in an edit. When exactly does the error occur? Is that after making some changes?

jxsvoboda:
I changed addr48_t to be a structure again (now eth_addr_t). That might hopefully help. If not, you can create a new PR (I will close this PR). However, what we haven't heard yet is how you came to the conclusion that the variable is not initialized. If this was found by some static checker, it's a false positive. The above mentioned change might have made it more clear and helped the checker understand the code. If not, you need to tell us which checker found this so that we can look into it. I don't think there is a bug in the code here.

PR #200, opened 2020-06-26
Display server

jxsvoboda:
Old Compositor
--------------
Why invest so much effort if we have a 'working' compositor?
  * messy, not well structured code, no unit tests, not testable
  * buffering policy dictated by protocol
      + required a lot of memory, no way to conserve when needed
  * fixed internal pixel format
      + each pixel needs to be individually converted to screen format
         by kfb. Can never be as fast as when compositing in native
         screen format (just copying data)
      + cannot support higher bit depths than RGB/888
      + cannot efficiently support lower bit depths

Fixing just a part of these issues would be more effort than starting
from scratch.

New Display Server
------------------
Cornerstone of a new graphics stack design
  + Graphic context is an abstract output/rendering API
  + Display server is a multiplexer (GC used on both input and output side)

More flexibility through clever abstraction
  + choose your buffering policy (zero/single/double)
  + choose where rendering occurs (client/server/display driver)
  + support different screen formats (<32bpp and >32 bpp) efficiently
  + possibly support HW blitters (accelerated indirect rendering)
  + preparation for multipoint support
  + possibility of multiple displays per display server instance

Testing
  + 140+ PCUT test cases
  + 1100+ individual PCUT assertions

The state of things right now
-----------------------------
Effort to be able to land new server as soon as possible
  * drop-in replacement for compositor
  * implement minimum features to be on-par with compositor

This means some things are not implemented or not final
  * running in same internal format as compositor (RGBA8888)
  * flicker-free single-buffered display not implemented
  * multipoint support not finished
  * graphical applications still have display service name as mandatory
    argument (soon to be changed)

What changes you will find as a user
------------------------------------
  * no more transparent, scaling or rotating windows
  * hid/display instead of comp:0/winreg
  * different background, arrow cursor
  * cursor changes when pointer is above resize areas
  * properly previewed minimum size of resized window
  * when you try to make window too small, it gets the correct minimum size

Future work on the DS
---------------------
Soon (T.M.):
  * natively support screen pixel format
      + faster
      + event faster and less memory usage on lower bit depths
      + requires support from the clients (GUI library) for maximum benefit
Other possible improvements:
  * mode setting
  * single-buffer flicker-free display
  * zero-buffer windows
  * multipoint (with RFB)

What's next?
------------
  * Next big step is implementing a proper user-interface library
    + libgui contains just 3 real widgets (button, label, canvas)
    + need to develop graphic primitives in tandem (e.g. line) as needed
    + support graphics and text mode


uspace/lib/congfx/meson.build
```diff
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2011 Petr Koupy
+# Copyright (c) 2015 Jiri Svoboda
```
jermar, line 2: 2015?

uspace/lib/ddev/src/ddev_srv.c
```diff
@@ -0,0 +1,165 @@
...
+ * for a DDF driver or a regular server. This would not be needed if we
+ * had a proper way of creating an endpoint and passing it to our client.
+ */
+static void ddev_get_gc_srv(ddev_srv_t *srv, ipc_call_t *icall)
+{
+	sysarg_t arg2;
+	sysarg_t arg3;
+	errno_t rc;
+
+	printf("ddev_get_gc_srv\n");
```
jermar, line 61: debug print

uspace/lib/ddev/src/ddev_srv.c
```diff
@@ -0,0 +1,165 @@
...
+	async_answer_2(icall, rc, arg2, arg3);
+}
+
+/** Get display device information */
+static void ddev_get_info_srv(ddev_srv_t *srv, ipc_call_t *icall)
+{
+	ddev_info_t info;
+	errno_t rc;
+
+	printf("ddev_get_info_srv\n");
```
jermar, line 78: debug print

jermar: Hm, there are more printfs in this file. Is this by intention or is this a left-over?

uspace/lib/gfx/private/bitmap.h
```diff
@@ -0,0 +1,56 @@
...
+
+/** @addtogroup libgfx
+ * @{
+ */
+/**
+ * @file Bitmap structure
+ *
+ */
+
+#ifndef _GFX_PRIVATE_COLOR_H
```
jermar, line 37: Wrong header guard. We should really be using #pragma once :-)

jxsvoboda: It should be possible to improve our tooling to detect this sort of issues.

uspace/lib/ddev/src/ddev_srv.c
jxsvoboda: Fixed.

uspace/lib/gfx/private/bitmap.h
jxsvoboda: Fixed

uspace/lib/ddev/src/ddev_srv.c
jxsvoboda: Fixed.

uspace/lib/congfx/meson.build
jxsvoboda: Fixed

PR #201, opened 2020-07-05
Enable window double buffering by default

jermar:
At this moment, this setting results in better user
experience on all platforms.

### DIFF
--- a/HelenOS.config
+++ b/HelenOS.config
@@ -611,7 +611,7 @@
 ! [CONFIG_FB=y] CONFIG_DISP_DOUBLE_BUF (y/n)
 
 % Window double buffering
-! [CONFIG_FB=y] CONFIG_WIN_DOUBLE_BUF (n/y)
+! [CONFIG_FB=y] CONFIG_WIN_DOUBLE_BUF (y/n)
 
 % Start AP processors by the loader
 ! [PLATFORM=sparc64&CONFIG_SMP=y] CONFIG_AP (y/n)


jermar:
I was thinking whether to make this the default in the light of the approaching release. If window double buffering is not set, I get the nasty flicker when moving windows.

jxsvoboda:
Not sure if we're seeing the same problem. When I start e.g. top in vterm so that the window gets updated regularly and then move any window around, the vterm window will flicker during updates - the background color will be visible momentarily in some areas. Same effect can sometimes be seen just when moving the mouse cursor (albeit it's much less pronounced). Now I tried enabling window double-buffer, but it had made no change for me. And it shouldn't. Window double-buffer should prevent you from seeing artefacts resulting from painting the window at a moment when it is partially drawn by libgui. But that's not what we're seeing is artifacts of display server's rendering of the display. Also note that window double-buffer will double the amount of memory required to store window compared to current state or compared to compositor (which always used single-buffered windows).

jermar:
Hm, I can't reproduce this now either. Only if I leave window double buffering off (default) and explicitly disable display double buffering am I able to get the flickering back when moving windows. So I don't know what happened when I first built this. It feels like something in the configuration got out of sync. I will keep an eye on similar behavior, but will close the PR now.

jxsvoboda:
I fixed the flicker issue in https://github.com/HelenOS/helenos/commit/d74a2b8558df383430ebac0c405222e9d82b5bef

PR #204, opened 2020-11-28
New user inteface library

jxsvoboda:
Replace GUI library libgui with new user interface library libui, which is based natively on GFX and using proportional font. We get rid of most of libdraw (except for TGA decoding -> libgfximage) and libsoftrend (except for pixel conversions -> libpixconv). A full suite of unit tests is included.

Other notable changes/improvements:
 - added library for editing and displaying proportional fonts, libgfxfont, simple font editor application
 - changed look
 - applications changed name (vterm-> Terminal (terminal), vdemo -> UI Demo (uidemo), vlaunch -> Launcher (launcher))
 - Calculator received more standard button layout and can be controlled using the keyboard
 - 'animated' buttons
 - viewer has a more useful fullscreen mode, activated with -f option

Caveats:
 - client-side rendering not implemented yet (always render server-side) and thus window double-buffer option has no effect
 - widgets positions and sizes are specified manually - automatic layout not implemented
 - window decoration is now 2 pixels higher, which means OCR (in helenos-ci) needs to be updated (see https://github.com/HelenOS/ci/pull/4)


uspace/lib/riff/include/types/riff/rwave.h
```diff
@@ -0,0 +1,103 @@
...
+ */
+
+/** @addtogroup libriff
+ * @{
+ */
+/**
+ * @file Waveform Audio File Format (WAVE) types.
+ */
+
+#ifndef RIFF_TYPES_WAVE_H
```
jermar, line 36: Strictly speaking, RIFF_TYPES_RWAVE_H?

jermar (commented):
So I have seen everything. I can't say I really reviewed it though. Good job and fire at will!

PR #206, opened 2020-12-09
added command line option for running multiple cpus

parth-27:


vhotspur:
Thanks for splitting the PR!

This looks much easier to review. Though I would prefer that the number of CPUs is a normal option (like `-cpu N`).

PR #207, opened 2020-12-19
grep command added

parth-27:


matthieuriolo:
Hey Parth

Thanks for your contribution! But please make sure to run ```make ccheck``` before committing as only well formated code will be accepted.

I briefly looked over your code and there are few things which stands out. 

1. You have two commits with the same message. Please merge those two together: ```git rebase -i HEAD^2```
2. Your regex implementation does not follow the notation of the regex at all... just some examples
```':a'   A colon matches a class of characters described by the following```
character classes are defined by using a backslash: [ \w](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Character_Classes) or the bracket notation [:alpha:]
```'-'    An expression followed by a minus sign optionally matches"```
this is done with a [question mark](https://regular-expressions.mobi/optional.html?wlr=1) in regex
3. The whole regex portion really should be in an external library. That way it's reusable. Considering how difficult regex parsing is, I really dont recommand you to start with this feature 😉 having grep without regex functionality but instead with the feature to do recursive search and negation would be a better start

Best regards and a lovely new year, Matt

jxsvoboda:
Hi Parth,

thanks for the contribution, although I would recommend that If you have an idea for contributing, you should first discuss it on the mailing list so we can flesh it out design-wise. Although current shell Bdsh looks quite UNIXy, I would refer you to http://www.helenos.org/ticket/653 there is a plan to replace it with a different, non-UNIX-like shell. In other words adding a typical UNIX command like grep is not a step in the right direction. On a lower level, I see quite a few issues, like the commenting style is all wrong. The grep.c has some very strange copyright. Is that a 3rd-party code (i.e. code that you haven't written yourself)? Where does it come from? We prefer original code.

All in all, I'm not sure if it is a good use of time to continue with this PR.

It's great that you are interested in contributing, but as I said before, it's good to discuss your contribution first and get advice from the core team to make sure you are headed in the right direction.

PR #213, opened 2021-06-19
Add a some more functions to the math library

maurizio-lombardi:
Hello guys,

To make the porting of the NES emulator possible, I needed a few math operations (log(), sqrt(), pow() and scalbn()) that  are not available in the mainline HelenOS tree.
Manuele copied part of the msun math library (BSD-like licensed, very permissive) and then I synced it to a recent version of FreeBSD src tree.

I am 100% aware that you don't like to merge code coming from other projects, yet this code was rotting in my private branch and I opened this pull request just to make you aware that some work has been done in this area. I'm not asking to merge it right now, it may become useful in future.

Note: I had to fix a "fix comparison between signed and unsigned integers" warning that prevented the code from successfully building. I think my fix is safe, but requires some careful review... because all this math code looks like black magic to me.

### DIFF
--- a/uspace/lib/math/generic/internal.h
+++ b/uspace/lib/math/generic/internal.h
@@ -30,6 +30,120 @@
 #ifndef MATH_INTERNAL_H_
 #define MATH_INTERNAL_H_
 
+#include <stdint.h>
+
+#if __BE__
+
+typedef union
+{
+	double value;
+	struct
+	{
+		uint32_t msw;
+		uint32_t lsw;
+	} parts;
+	uint64_t word;
+} ieee_double_shape_type;
+
+#endif
+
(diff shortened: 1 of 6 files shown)


jxsvoboda:
Hi Maurizio! Thanks! pow(), log() and sqrt() were available in HelenOS until they were unceremoniuously ripped out in changeset https://github.com/HelenOS/helenos/commit/516e7808c5e1e23b10a77bb1429054a16f6fedc3
I think we might try to revive the old code (while adding some rreasonable tests and making sure it's sane). I'd prefer having a say less efficient and reasonable accurate, but understandable version that one that is super optimized and accurate, but obscure.
Cheers,
Jiri

PR #214, opened 2021-06-22
Graphics improvements for the Raspberry PI

maurizio-lombardi:
Currently, the SPARTAN kernel assumes a display size of 640x480

With this patchset, it will read the correct size from VideoCore,
looks much better on large screens.

http://bsdbackstore.it/misc/IMG_3463.jpg


jxsvoboda:
Nice work!

PR #215, opened 2021-06-23
Configure NS16550 transmission format settings on initialization.

dusxmt:
Currently, the NS116550 serial line controller is left with its settings as it was left by the boot firmware and/or bootloader.  On my computer, this was an invalid configuration, and it left me with a really slow booting system, since each output character had to go through the full timeout loop in ns16550_sendb().

This patch adds the necessary bit and register descriptions to configure the baud rate and transmission settings, as well as configuring them on initialization, currently with values matching the ns8250 userspace character device driver (38400 baud, 8-bit words, 2 stop bits, no parity).

This could perhaps be changed to be adjustable with a kernel command-line argument, or through the configuration system.

This change does not affect emulators, since those largely ignore these settings.

### DIFF
--- a/kernel/arch/amd64/src/amd64.c
+++ b/kernel/arch/amd64/src/amd64.c
@@ -223,6 +223,8 @@ void amd64_post_smp_init(void)
 	    ns16550_init(NS16550_BASE, 0, IRQ_NS16550, NULL, NULL,
 	    ns16550_out_ptr);
 	if (ns16550_instance) {
+		ns16550_format_set(ns16550_instance, 38400,
+		    LCR_PARITY_NONE | LCR_STOP_BIT_TWO | LCR_WORD_LEN_8);
 #ifdef CONFIG_NS16550
 		srln_instance_t *srln_instance = srln_init();
 		if (srln_instance) {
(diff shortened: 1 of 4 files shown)


dusxmt:
Well, I tried adding options into 'ninja config' to configure the transmission settings and/or leave them unconfigured like it was by default, but as I was testing out various combinations, I found out that there's actually a second serial driver, in userspace, ns8250, which would then override my settings and garble up my text output.

I guess this partially confirms to me that this is a bug, since the settings are set later in the boot process, whatever was left there by the boot firmware will be overridden once the system loads up the userspace driver, so it only makes sense to configure those settings as early as possible, to allow the user to get a full boot log via serial - from both parts of the boot process.

jxsvoboda:
Hi!

Thanks for the PR! It is really well and nicely written. I found out only one small, but important issue. ns16550 is used on different platforms, not just PC and not just for general-purpose serial ports. On SPARC Ultra, for example, the keyboard is connected via ns16550. Changing the communication parameters there would probably render the keyboard inoperable.

Instead of setting

dusxmt:
Hey there!

Thanks for the input, I changed it like you requested, tested it on real hardware on ia32.

jxsvoboda:
Good work! Thank you!

PR #217, opened 2021-08-16
Full support pipes

manux81:
These changes start a full pipes support.
In particular this branch introduces following commands:
- **>** Save output to a file.
- **<** Read input from a file.
- **|** Send the output from one program as input to another program.

### DIFF
--- a/uspace/app/bdsh/input.c
+++ b/uspace/app/bdsh/input.c
@@ -226,10 +226,16 @@ static errno_t process_input_nohup(cliuser_t *usr, list_t *alias_hups, size_t co
 		default:
 			break;
 		}
+
 		if (pipe_count > MAX_PIPES) {
 			rc = ENOTSUP;
 			goto finit;
 		}
+	}
+
+	if (wait_from || wait_to) {
+		printf("Parse error near `\\n'\n");
+		goto finit;
 	}
 
 	unsigned int cmd_token_start = 0;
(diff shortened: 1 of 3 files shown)


jxsvoboda:
I'm not sure if it's worth spending much time with Bdsh which is planned to be replaced, but it would make sense if you would be able to sort out the potential VFS issues that (supposedly) prevent UNIX-style pipes (not DOS-style pipes) to be implemented.

If I understand correctly, you have implemented so far a sort of DOS-style pipe, where the first command is executed, the output is sent to /tmp/pipe, then the second command is executed with input redirected from /tmp/pipe, etc?

How did you test it? The only command I could think of that can be reasonably used as a filter is "cat -s -". I tried it with < and it works (as bad as before with from x|cat -s -) but with "cat /demo.txt | cat -s -" I get an error "cannot open -". Also I would note that it would be nice to delete the temporary file. And the temporary file would be ideally created with tmpfile(), but if you need a name, you could at least generate it with tmpnam().

manux81:
Below you can find my tests:
1- > Save output to a file:
1A) ls > prova.txt
1B) alias pluto='ls'
pluto > prova.txt
1C) ls > 
>> Print a parser error near \n
2-> Send the output from one program as input to another program.
I create a program like follow
int main() {
  while (!feof(stdin)) {
    size_t size = fread(str, 1, 255, stdin); // read from stdin
    str[254] = 0;
    fwrite(str, 1, size, stdout); // print read content out to stdout
 }
 return (0);
}

cat demo.txt | in2out

3) < Read input from a file.
No already test yet.

Actually I'm stopping test because I found a error in bdsh that it not depends by my code.
If  you type follow commands bdsh stops working:
help | to prova.txt
cat prova.txt

manux81:
With last version I checked follow example and it seems work:
cat demo.txt | cat -s -
it works with my test too:
cat demo.txt | in2out

manux81:
Hi @jxsvoboda, I done changes that you ask to me.

PR #220, opened 2022-01-31
Enable -march for uspace again

jermar:


jermar:
Already fixed by another commit.

PR #222, opened 2022-08-15
Simplify timeout implementation in kernel

le-jzr:
Replaces countdown with an immutable deadline, which allows us to remove `timeout->lock` and simplify the code quite a bit.

### DIFF
--- a/kernel/generic/src/time/clock.c
+++ b/kernel/generic/src/time/clock.c
@@ -162,9 +162,7 @@ void clock(void)
 			timeout_t *timeout = list_get_instance(cur, timeout_t,
 			    link);
 
-			irq_spinlock_lock(&timeout->lock, false);
 			if (current_clock_tick <= timeout->deadline) {
-				irq_spinlock_unlock(&timeout->lock, false);
 				break;
 			}
 
@@ -172,7 +170,6 @@ void clock(void)
 			timeout_handler_t handler = timeout->handler;
 			void *arg = timeout->arg;
 
-			irq_spinlock_unlock(&timeout->lock, false);
 			irq_spinlock_unlock(&CPU->timeoutlock, false);
 
 			handler(arg);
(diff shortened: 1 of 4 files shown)


kernel/generic/src/time/clock.c
```diff
@@ -158,7 +152,7 @@ void clock(void)
 	size_t i;
 	for (i = 0; i <= missed_clock_ticks; i++) {
 		/* Update counters and accounting */
```
jermar: This is now done on line 142...

kernel/generic/src/time/timeout.c
```diff
@@ -103,9 +88,6 @@ void timeout_register(timeout_t *timeout, uint64_t time,
 	irq_spinlock_lock(&CPU->timeoutlock, true);
 	irq_spinlock_lock(&timeout->lock, false);
 
-	if (timeout->cpu)
```
jermar, line 106: If I understand this correctly, we are giving up the possibility to detect that a timeout is being registered for the second time here...

kernel/generic/src/time/timeout.c
```diff
@@ -86,32 +86,29 @@ void timeout_register(timeout_t *timeout, uint64_t time,
 {
 	irq_spinlock_lock(&CPU->timeoutlock, true);
 
+	assert(!link_in_use(&timeout->link));
```
jermar, line 90: I guess this explains my previous comment.

kernel/generic/src/time/clock.c
```diff
@@ -149,37 +151,29 @@ void clock(void)
 	 * run all expired timeouts as you visit them.
 	 *
 	 */
-	size_t i;
-	for (i = 0; i <= missed_clock_ticks; i++) {
-		/* Update counters and accounting */
```
jermar, line 160: Okay, also resolves my previous comment.

jermar (approved):
Looks good to me. Sorry for the looong delay. Reading the commits from oldest to newest, I had some comments/questions, but the newer commits resolved them, so I think this is good to go.

PR #223, opened 2022-08-22
Fix partial release of first extent

hekkihek:
For partial release of first extent, first_fblock > start of extent, this fix makes delete_count smaller than block_count instead of larger as it should be.


le-jzr:
Indeed! Nice catch, thank you. :)

PR #225, opened 2023-04-18
Fixed warning

polluks:
WARNING: Running the setup command as `meson [options]` instead of `meson setup [options]` is ambiguous and deprecated.

### DIFF
--- a/configure.sh
+++ b/configure.sh
@@ -163,7 +163,7 @@ fi
 
 sed "s:@COMPPREFIX@:$compprefix:g" "$cross_def" > crossfile || exit 1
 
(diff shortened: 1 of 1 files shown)


le-jzr:
Hello, thank you :)
Before I merge this, could you check if this affects the least Meson version we can use? We might want to update wiki if we need a recent version for any deprecation fixes.

polluks:
It's backward compatible anyway, see https://lists.buildroot.org/pipermail/buildroot/2022-December/657900.html

le-jzr:
Completely forgot about this, sorry about that. Thank you for the fix. :)

PR #227, opened 2023-10-22
Deduplicate identical enums for window resizing

le-jzr:
A newer version of GCC appropriately warns about different enum types being mixed together.
Instead of an explicit typecast, which only hides the underlying issue, I'm thinking there really should not be two identical enums being used like this.

This patch removes `ui_wdecor_rsztype_t` and uses `display_wnd_rsztype_t` in its place.
@jxsvoboda this is your wheelhouse. What do you think?

### DIFF
--- a/uspace/lib/display/include/types/display/wndresize.h
+++ b/uspace/lib/display/include/types/display/wndresize.h
@@ -37,6 +37,8 @@
 
 /** Window resize type */
 typedef enum {
+	display_wr_none = 0,
+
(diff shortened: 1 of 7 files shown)


jxsvoboda:
I don't want to mix in the display server interface into UI interface. The UI interface should fully isolate you from what's beneath it (it needn't be display server, it can be console or something else altogether). I think the proper way to solve this is to properly convert between the two types, even if they are similar.

PR #228, opened 2023-11-13
Updates for MSIM >2.0.0

vhotspur:
I assume we do not need to keep backward compatibilty and support both pre and post 2.0 versions of MSIM.


vhotspur:
Ping :-). Any objections, please?

PR #229, opened 2023-12-15
Packet capture prototype

boba-buba:
## Ticket 629 (http://www.helenos.org/ticket/629).

That is a prototype of the support for network packet capture.  I am doing it as a bachelor thesis at MFF UK. I will be grateful for any kind of feedback.

The support is implemented for e1k driver. And I tested it on ia32 architecture mostly. Now there are only two requests supported: start dumping all packets, stop dumping.

## Usage
Application: pcapctl

pcapctl start <file name> (Example: `pcapctl start data.pcap `, file format is .pcap)

pcapctl stop

If someone is trying to start dumping, when it is already started the original file will be closed and saved, new file will be created and packets will be dumped to new file. If trying to stop when already stopped nothing will happen. 

Then after getting the file out of HelenOS (through websrv for example)  it can be opened  with WireShark to see the packets, that were dumped.

## Implementation

During start of the system `pcap` category created and e1k driver is added to `pcap` category.  User uses application `pcapctl`.  `pcapctl start data.pcap` will send request for start of the dumping. Location server will forward it to the first device in category pcap. And driver will start dumping packets. For stopping it is the same.


vhotspur:
I have switched the PR into the ready state because I think it is in good enough shape to be tried and tested :-).

BTW, the following commands can be used to capture our webserver traffic:

```
pcapctl start capture.pcap
webserv
# From host, open localhost:8080 and browse the pages
pcapctl stop
cp capture.pcap data/web
```

And on host we can open it in Wireshark.

```
wget "http://localhost:8080/capture.pcap"
wireshark capture.pcap
```

The plan is now to write the thesis text but I think it would be really nice if Nataliia can get some feedback from the rest of the community too (and not only my complaints).

uspace/lib/pcap/src/pcap_iface.c
```diff
@@ -0,0 +1,107 @@
...
+
+errno_t pcap_init(const char *name)
+{
+	errno_t rc = pcap_writer_to_file_init(&pcap_writer, name);
+	return rc;
+}
+
+void pcap_add_packet(const void *data, size_t size)
+{
+	if (&pcap_writer.data == NULL)
```
vhotspur, line 79: Looks like a typo: the `&` should not be here.

jxsvoboda (changes requested):
Looks pretty good. I added some comments, most of them pretty minor. I really wonder if we can avoid repeating the same code in all nic drivers. Why not do it in libnic, any particular reason?

PR #230, opened 2024-01-05
Add export-dev target to export all libraries

vhotspur:
The new target `export-dev` builds all libraries and copies the created `.a` archives and library headers to `export-dev/` subdirectory.

Unlike the existing approach from `tools/export.sh` we do not list the libraries manually but copy all of them automatically. This simplifies maintenance when new library is added and it should also simplify coastline builds as no change to HelenOS tree would be needed when the ported software needs another library not yet mentioned in `export.sh` (I encountered the issue when trying to fix the build of MSIM).

The approach relies as much on Meson to do the heavy-lifting as possible and the invoked shell script merely copies the files (seems that Meson is not able to do that by itself inside a target). The script honors exported `DESTDIR` variable if set and ensures thin archives are converted before export.

For normal libraries having standard `include/` subdirectory, everything happens automatically. For `libc` and `libposix` the include paths need to be added manually (unless someone knows how to expand `include_directories` to an array of paths (strings)).

Usage:

```
# Exports into export-dev/ with
# lib/ and include/ subdirectories
ninja export-dev

# Exported into harbour-export directory.
env DESTDIR=harbour-export ninja export-dev
```

I would be glad if someone who actually understands Meson would have a look if the chosen approach is reasonable. Thanks!


le-jzr:
Seems reasonable to me. I'm very much in favor of simplifying things this way.

vhotspur:
> Seems reasonable to me. 

Thanks for checking it!

I pushed some smaller fixes and also rewritten `helenos-pkg-config` to Python to (hopefully) make it more maintainable.

With these fixes I can again compile all harbours :-)

I plan to merge this next week.

PR #231, opened 2024-06-05
Change internal UUID representation and use correct GPT UUID endianness

mcimerman:
### Representation part
Change the UUID internal representation so that it has named fields.

### GPT part
Encode the partition type UUID in little-endian format so that the type of GPT partition created within HelenOS can be recognized outside of HelenOS.

I haven't found anything regarding the byte order in the UEFI spec, but this is the way it is done.

sections[1]:
 - time_low, time_mid, time_hi_and_version are stored in little-endian
 - clock_seq_hi_and_reserved, clock_seq_low, node are stored in big-endian.

Actually it is all little-endian, but because of missing hyphens it appears that the second half is big-endian. [2] [3]

[1] https://www.rfc-editor.org/rfc/rfc4122#section-4.1.2
[2] https://en.wikipedia.org/wiki/Universally_unique_identifier#Encoding
[3] https://devblogs.microsoft.com/oldnewthing/20220928-00/?p=107221

### DIFF
--- a/uspace/lib/c/generic/rndgen.c
+++ b/uspace/lib/c/generic/rndgen.c
@@ -91,6 +91,32 @@ errno_t rndgen_uint8(rndgen_t *rndgen, uint8_t *rb)
 	return EOK;
 }
 
+/** Generate random 16-bit integer.
+ *
+ * @param rndgen Random number generator
+ * @param rw Place to store random 16-bit integer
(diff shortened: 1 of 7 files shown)


mcimerman:
I also added helper function to be more compact.

jxsvoboda:
Thanks.

PR #234, opened 2024-06-29
Minimize an active window when its entry is clicked in the taskbar

MaaSTaaR:


uspace/app/taskbar/wndlist.c
```diff
@@ -716,13 +716,27 @@ errno_t wndlist_repaint(wndlist_t *wndlist)
...
 
 	/* ID of device that clicked the button */
 	dev_id = entry->wndlist->ev_idev_id;
 
-	(void) wndmgt_activate_window(entry->wndlist->wndmgt,
-	    dev_id, entry->wnd_id);
+	rc = wndmgt_get_window_info(entry->wndlist->wndmgt,
+	    entry->wnd_id, &winfo);
+	if (rc == EOK) {
+		minimized = (winfo->flags & wndf_minimized) == 0;
```
jxsvoboda, line 730: Note that the logic is reversed w.r.t. the name of the variable: minimized will be true iff the window is not minimized

uspace/app/taskbar/wndlist.c
```diff
@@ -716,13 +716,27 @@ errno_t wndlist_repaint(wndlist_t *wndlist)
...
+	}
+
+	if (!minimized) {
+		(void) wndmgt_activate_window(entry->wndlist->wndmgt,
+	        dev_id, entry->wnd_id);
+	} else {
```
jxsvoboda, line 736: Now here's the big problem. If you click the button of a window that's not minimized and is *not focused*, instead of activating the window it will minimize it. Also note that with multiple seats, each window can have zero, one or more foci. A window should only be minimized if it is the focus of the same seat that clicked the window button.

uspace/app/taskbar/wndlist.c
```diff
@@ -716,13 +716,27 @@ errno_t wndlist_repaint(wndlist_t *wndlist)
...
+		(void) wndmgt_deactivate_window(entry->wndlist->wndmgt,
```
jxsvoboda, line 737: I think the operation should simply be called 'minimize', because that's what it does (this pertains to all other places too).

uspace/app/taskbar/wndlist.c
```diff
@@ -716,13 +716,27 @@ errno_t wndlist_repaint(wndlist_t *wndlist)
...
-	    dev_id, entry->wnd_id);
+	rc = wndmgt_get_window_info(entry->wndlist->wndmgt,
+	    entry->wnd_id, &winfo);
+	if (rc == EOK) {
+		minimized = (winfo->flags & wndf_minimized) == 0;
+	}
+
+	if (!minimized) {
+		(void) wndmgt_activate_window(entry->wndlist->wndmgt,
+	        dev_id, entry->wnd_id);
```
jxsvoboda, line 735: The indentation is incorrect. Did you run 'make ccheck'?

jxsvoboda (changes requested):

uspace/lib/wndmgt/include/wndmgt.h
```diff
@@ -47,6 +47,7 @@ extern errno_t wndmgt_get_window_info(wndmgt_t *, sysarg_t,
     wndmgt_window_info_t **);
 extern void wndmgt_free_window_info(wndmgt_window_info_t *);
 extern errno_t wndmgt_activate_window(wndmgt_t *, sysarg_t, sysarg_t);
+extern errno_t wndmgt_deactivate_window(wndmgt_t *, sysarg_t, sysarg_t);
```
jxsvoboda, line 50: For this new operation unit tests should be added to /data/helenos/master/uspace/lib/wndmgt/test/wndmgt.c, similar to PCUT_TEST(activate_window_failure) and PCUT_TEST(activate_window_success).

PR #237, opened 2024-11-10
ata_channel_destroy(): free the channel

mcimerman:
While browsing the ATA code again, I noticed that the initial call to calloc() doesn't have a free() pair.

### DIFF
--- a/uspace/lib/ata/src/ata.c
+++ b/uspace/lib/ata/src/ata.c
@@ -259,12 +259,14 @@ errno_t ata_channel_destroy(ata_channel_t *chan)
 		rc = ata_device_remove(&chan->device[i]);
 		if (rc != EOK) {
 			ata_msg_error(chan, "Unable to remove device %d.", i);
-			break;
+			fibril_mutex_unlock(&chan->lock);
+			return rc;
 		}
 	}
 
 	ata_bd_fini_irq(chan);
 	fibril_mutex_unlock(&chan->lock);
+	free(chan);
 
 	return rc;
 }


mcimerman:
Looking back at this PR, I think we should free() the channel only when `rc == EOK`.

What do you think?

mcimerman:
I think this seems more reasonable. After my previous PR #236, `ata_bd_fini_irq(chan)` is called even though there was an error, and the current PR is wrong again (due to the `break`), as it frees the channel even when an error occurs.

Here it is fixed, let me know if it looks ok and I will force push to my branch so you can pull it:

```C
/** Destroy ATA channel. */
errno_t ata_channel_destroy(ata_channel_t *chan)
{
	int i;
	errno_t rc;

	ata_msg_debug(chan, ": ata_channel_destroy()");

	fibril_mutex_lock(&chan->lock);

	for (i = 0; i < MAX_DEVICES; i++) {
		rc = ata_device_remove(&chan->device[i]);
		if (rc != EOK) {
			ata_msg_error(chan, "Unable to remove device %d.", i);
			fibril_mutex_unlock(&chan->lock);
			return rc;
		}
	}

	ata_bd_fini_irq(chan);
	fibril_mutex_unlock(&chan->lock);
	free(chan);

	return rc;
}
```

zmmaj:
When will be this usable, since HelenOS did not see physical hard disc... Ide do not work Sata, same..

le-jzr:
> I think this seems more reasonable. After my previous PR #236, `ata_bd_fini_irq(chan)` is called even though there was an error, and the current PR is wrong again (due to the `break`), as it frees the channel even when an error occurs.

I looked at what's happening when error is returned from the function, and it seems to me that the whole error path is just underspecified and inconsistently handled. Right now, not freeing on error seems like the safe choice (there might be leaks in some circumstances, but no crashes). Though eventually, someone should probably look through the entire driver stack and better specify what should happen in error cases, because the current handling doesn't feel right to me.

le-jzr:
LGTM. Thanks!

PR #238, opened 2024-12-22
Allow bitmap to be used in userspace

mcimerman:
Hi,

I happen to need to use a bitmap for my project, but the one in HelenOS is
not available in libc, only for the kernel...

So I included it to be built in libc, but then userspace doesn't see <[typedefs.h](https://github.com/HelenOS/helenos/blob/master/kernel/generic/include/typedefs.h)>
kernel header, which seems to not be needed by the bitmap.

Objections? Ok?

### DIFF
--- a/common/adt/bitmap.c
+++ b/common/adt/bitmap.c
@@ -26,7 +26,7 @@
  * THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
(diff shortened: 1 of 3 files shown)


vhotspur:
LGTM.

BTW, we have a nice dichotomy there as `odict` is also shared in kernel and userland and while it has guard `_LIBC_ODICT_H_` it belongs to `kernel_generic_adt` documentation.

PR #239, opened 2025-01-16
Capability/kobject modifications

le-jzr:


kernel/generic/src/ipc/irq.c
```diff
@@ -294,15 +294,15 @@ static void irq_hash_out(irq_t *irq)
 	irq_spinlock_unlock(&irq_uspace_hash_table_lock, true);
 }
 
-static void irq_destroy(void *arg)
+static void irq_destroy(kobject_t *arg)
 {
-	irq_t *irq = (irq_t *) arg;
+	irq_kobject_t *kobj = (irq_kobject_t *) arg;
```
jermar, line 299: For the sake of consistency, irq_from_kobject(arg)?

jermar: I think also in other places...

kernel/generic/src/ipc/ipc.c
```diff
@@ -891,6 +883,8 @@ void ipc_init(void)
 	    NULL, 0);
 	answerbox_cache = slab_cache_create("answerbox_t", sizeof(answerbox_t),
 	    0, NULL, NULL, 0);
+	irq_cache = slab_cache_create("irq_t", sizeof(irq_kobject_t),
+	    0, NULL, NULL, 0);
```
jermar, line 887: Hm, so irq_cache is now a cache of irq_kobject_t objects. Wasn't it the case that before there were also pure irq_t objects not visible by userspace and therefore not referencable by a capability?

kernel/generic/src/ipc/irq.c
le-jzr: Yeah, probably. I omitted it here because it's all in the same file, but it makes sense to make it consistent.

kernel/generic/src/ipc/ipc.c
le-jzr: Yeah, they still exist. They just never used this slab for allocations. I think they are all just static.

le-jzr: I admit it's mildly confusing, I should probably rename it.

PR #240, opened 2025-01-25
fix calculation of TLS size for main fibril

mvolfik:
Before this patch, `_tcb_data_offset` always used `__progsymbols.elfstart`. However, that is wrong when it is being called from the loader server!! This can leave much less space for the TLS data than requested and TLS accesses will corrupt some memory!

Now we pass to it a pointer to the correct ELF, falling back to elfstart in the public tls_get call.

Also added some comments, explaining some confusing parts...

### DIFF
--- a/uspace/lib/c/generic/thread/tls.c
+++ b/uspace/lib/c/generic/thread/tls.c
@@ -84,6 +84,10 @@ void *tls_get(void)
 
 static tcb_t *tls_make_generic(const void *elf, void *(*alloc)(size_t, size_t))
 {
+	/*
+	 * See also rtld/module.c -> modules_process_tls(), where we have less
+	 * messy code for the dynamic-linking version of this.
+	 */
 	assert(!elf_get_phdr(elf, PT_DYNAMIC));
 #ifdef CONFIG_RTLD
 	assert(runtime_env == NULL);
@@ -126,6 +130,18 @@ static tcb_t *tls_make_generic(const void *elf, void *(*alloc)(size_t, size_t))
 		return tcb;
 
 	uintptr_t bias = elf_get_bias(elf);
+
+	/*
+	 * FIXME: I couldn't convince myself this is correct, but I couldn't
+	 * find a case where it breaks either: if the actual alloc_size is
+	 * bigger than the requested size (tls->p_memsz), the alignment padding
+	 * will be placed at the beginning (because TLS is (at least sometimes?)
+	 * indexed with negative offsets from the TCB pointer).
+	 *
+	 * Now we will copy the initialization data to a position at the start of
+	 * the allocation, so if the padding has nonzero size, if think the initialization
+	 * data is now incorrectly offset by its size?
+	 */
 
 	/* Copy thread local data from the initialization image. */
 	memcpy(data, (void *)(tls->p_vaddr + bias), tls->p_filesz);
--- a/uspace/lib/c/generic/thread/tls.c
+++ b/uspace/lib/c/generic/thread/tls.c
@@ -103,6 +103,13 @@ static tcb_t *tls_make_generic(const void *elf, void *(*alloc)(size_t, size_t))
 	 */
 	assert(tls_align <= PAGE_SIZE);
 
+	/*
+	 * FIXME: the calculation of alloc_size shouldn't include the alignment
+	 * of tcb_t (at least in Variant II)
+	 * See https://github.com/HelenOS/helenos/pull/240/files/4ef27ebf98a0656e09889b7d00efdec03343f1aa#r1929592924
+	 * (you will also need to fix _tcb_data_offset)
+	 */
+
 #ifdef CONFIG_TLS_VARIANT_1
 	size_t alloc_size =
 	    ALIGN_UP(sizeof(tcb_t), tls_align) + tls_size;
@@ -130,26 +137,6 @@ static tcb_t *tls_make_generic(const void *elf, void *(*alloc)(size_t, size_t))
 		return tcb;
 
 	uintptr_t bias = elf_get_bias(elf);
-
-	/*
-	 * FIXME: I couldn't convince myself this is correct, but I couldn't
-	 * find a case where it breaks either: if the actual alloc_size is
-	 * bigger than the requested size (tls->p_memsz), the alignment padding
-	 * will be placed at the beginning (because TLS is (at least sometimes?)
-	 * indexed with negative offsets from the TCB pointer).
-	 *
-	 * Now we will copy the initialization data to a position at the start of
-	 * the allocation, so if the padding has nonzero size, if think the initialization
-	 * data is now incorrectly offset by its size?
-	 *
-	 * Maybe a diagram helps explaining this?
-	 * |    allocation   |     |
-	 * | paddding | data | tcb |
-	 *   ^
-	 *   +--- we will copy the initialization data here
-	 *              ^
-	 *              +--- but the data should be actually here?
-	 */
 
 	/* Copy thread local data from the initialization image. */
 	memcpy(data, (void *)(tls->p_vaddr + bias), tls->p_filesz);
--- a/uspace/lib/c/generic/thread/tls.c
+++ b/uspace/lib/c/generic/thread/tls.c
@@ -58,10 +58,10 @@
 #error Unknown TLS variant.
 #endif
 
-static ptrdiff_t _tcb_data_offset(void)
+static ptrdiff_t _tcb_data_offset(const void* elf)
 {
 	const elf_segment_header_t *tls =
-	    elf_get_phdr(__progsymbols.elfstart, PT_TLS);
+	    elf_get_phdr(elf, PT_TLS);
 
 	size_t tls_align = tls ? tls->p_align : 1;
 
@@ -79,7 +79,7 @@ void *tls_get(void)
 #ifdef CONFIG_RTLD
 	assert(runtime_env == NULL);
 #endif
-	return (uint8_t *)__tcb_get() + _tcb_data_offset();
+	return (uint8_t *)__tcb_get() + _tcb_data_offset(__progsymbols.elfstart);
 }
 
 static tcb_t *tls_make_generic(const void *elf, void *(*alloc)(size_t, size_t))
@@ -113,11 +113,11 @@ static tcb_t *tls_make_generic(const void *elf, void *(*alloc)(size_t, size_t))
 
 #ifdef CONFIG_TLS_VARIANT_1
 	tcb_t *tcb = area;
-	uint8_t *data = (uint8_t *)tcb + _tcb_data_offset();
+	uint8_t *data = (uint8_t *)tcb + _tcb_data_offset(elf);
 	memset(tcb, 0, sizeof(*tcb));
 #else
 	uint8_t *data = area;
-	tcb_t *tcb = (tcb_t *) (data - _tcb_data_offset());
+	tcb_t *tcb = (tcb_t *) (data - _tcb_data_offset(elf));
 	memset(tcb, 0, sizeof(tcb_t));
 	tcb->self = tcb;
 #endif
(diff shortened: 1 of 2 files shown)


mvolfik:
So I did some more tests on the TLS size / alignment confusion, and (at least on IA-32) it seems to behave correctly: apparently the compiler expects the TLS size to be rounded up to the alignment, and generates accesses accordingly. For this code:

<details>
<summary>Code (tap to embiggen)</summary>

```c
#include <align.h>
#include <fibril.h>
#include <stdalign.h>
#include <stdio.h>
#include <tls.h>

static fibril_local _Alignas(128) int var1 = {1};
static fibril_local char var2[4] = {'a', 'a', 'a', 'a'};

__attribute__((noinline)) static void print_var1() {
  printf("var1: %x\n", var1);
}
__attribute__((noinline)) static void print_var2() {
  for (int i = 0; i < 4; i++) {
    printf("%c", var2[i]);
  }
  printf("\n");
}

int main(int argc, char **argv) {
  printf("tls addr: %p\n", __tcb_get());
  printf("addr of var1: %p\n", &var1);
  printf("addr of var2: %p\n", &var2);
  print_var1();
  print_var2();

  var1 = 2;
  var2[3] = 'd';

  print_var1();
  print_var2();
}
```

</details>

We get this access to var2

```
 80481bd:	65 a1 00 00 00 00    	mov    %gs:0x0,%eax
 80481c3:	8d b0 00 ff ff ff    	lea    -0x100(%eax),%esi
```

And these ELF headers:

```
Program Headers:
  Type           Offset   VirtAddr   PhysAddr   FileSiz MemSiz  Flg Align
  LOAD           0x000000 0x08048000 0x08048000 0x105c0 0x105c0 R E 0x1000
  LOAD           0x010600 0x08059600 0x08059600 0x002d8 0x1196c RW  0x1000
  TLS            0x010600 0x08059600 0x08059600 0x00084 0x00094 R   0x80
```

So although the requested TLS mem size is 0x94 (148 B), the access goes to %gs-0x100 (-256 B). Which is honestly quite weird, but well, we're dealing with a compiler and ELF ABI.

I'm not sure what's the best way to document this knowledge in the code. We might just link this thread in place of the current 'fixme' comment?

---

Oh, and by the way, here's some print of memory that we shouldn't access, that you currently get on master, when you make the `var2` array a bit longer in the code above:

`static fibril_local char var2[256] = {'a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a','a'};`

image

After this PR, you correctly get `aaaa...`. I guess this could be used to write some tests?

uspace/lib/c/generic/thread/tls.c
```diff
@@ -127,6 +131,26 @@ static tcb_t *tls_make_generic(const void *elf, void *(*alloc)(size_t, size_t))
...
+	 *
+	 * Maybe a diagram helps explaining this?
+	 * |    allocation   |     |
+	 * | paddding | data | tcb |
+	 *   ^
+	 *   +--- we will copy the initialization data here
+	 *              ^
+	 *              +--- but the data should be actually here?
+	 */
+
```
le-jzr, line 153: The diagram is incorrect. The padding of the TLS section in variant 2 is between data and tcb. This more or less directly follows from the offset calculations in [https://www.akkadia.org/drepper/tls.pdf](https://www.akkadia.org/drepper/tls.pdf).

le-jzr: Also, just common sense. Padding exists to make things aligned. `data` needs to be aligned to TLS alignment, and TCB needs to be aligned to whatever TCB alignment is.

So, in variant II, you allocate memory aligned to `tls_align` and `alignof(tcb_t)`, and add just enough padding that the start of `tcb_t` is aligned correctly within the allocated memory.

le-jzr: Now that I'm looking at it, I actually think the calculation in `modules_process_tls()` is incorrect. The old comment over it definitely is.

mvolfik: aha, welp, I based a lot of my understanding on the `modules_process_tls` comment :D if that is wrong, then I guess I'm wrong on many things...

mvolfik: > So, in variant II, you allocate memory aligned to both tls_align and TCB alignment, and add just enough padding after TLS data that the start of tcb_t is aligned correctly within the allocated memory.

but the accesses to TLS are made as offsets from the `tcb_t` pointer, right? so if we _"add just enough padding after TLS data"_ to align this pointer to whatever we want, then this could break the alignment of the data in TLS, no?

le-jzr: Actually, don't mind those comments, it seems I was actually wrong about that :D
Just reading through that drepper spec again.

le-jzr: Still working through this in my head, but if there is an issue there, it will only manifest if a module at an earlier address in memory has lower alignment than a module at a higher address. Since executable module is always #1, a way to demonstrate any such issue would be to make the main executable's TLS section have some ridiculously high alignment, and see if that borked up libc's TLS data.

le-jzr: For reference, these are offsets for variant II from the spec, but they are subtracted from the thread pointer rather than added.
```
tlsoffset(1) = round(tlssize(1), align(1))
tlsoffset(m+1) = round(tlsoffset(m) + tlssize(m+1), align(m+1))
````

Comment (mvolfik) [reply uspace/lib/c/generic/thread/tls.c:153]: note that this code path is (I think) only for statically linked executables, so I think what you are describing isn't applicable here (at least I think so)

Comment (le-jzr) [reply uspace/lib/c/generic/thread/tls.c:153]: > a way to demonstrate any such issue would be to make the main executable's TLS section have some ridiculously high alignment, and see if that borked up libc's TLS data.

Oh wait, this function is for statically linked executables only. So that's not an issue, in this case there is only one module and there can never be any padding before it.

Comment (le-jzr) [reply uspace/lib/c/generic/thread/tls.c:153]: So, the actual answer that the comment in rtld is correct, but only applies when we have a bunch of dynamic library modules mixed in. Even though there might be an extra padding at the start of the TLS allocation, each individual module's padding is after its data.

Comment (le-jzr) [reply uspace/lib/c/generic/thread/tls.c:153]: Actually, it might still be an issue if the static executable's TLS section is less aligned than a single pointer. Maybe try with TLS that only contains a single `char`, see if the executable gets linked with alignment of 1.

Comment (mvolfik) [reply uspace/lib/c/generic/thread/tls.c:153]: yeah, I'm still not convinced. Is the alignment of `tcb_t` a "public interface"? As in, can the OS arbitrarily change it, without it being an ABI change?

Comment (mvolfik) [reply uspace/lib/c/generic/thread/tls.c:153]: > Maybe try with TLS that only contains a single char, see if the executable gets linked with alignment of 1.

Unfortunately that won't work - multiple things somewhere in `libc` also use TLS, so even though my userspace program uses a single char, the final static binary contains more TLS data than that.

Comment (le-jzr) [reply uspace/lib/c/generic/thread/tls.c:153]: > yeah, I'm still not convinced. Is the alignment of `tcb_t` a "public interface"? As in, can the OS arbitrarily change it, without it being an ABI change?

See, that's why my original message was wrong. The scheme I described would work perfectly if TCB alignment was part of ABI, but it's not. The way offsets are calculated in variant II is independent of TCB alignment, so that would mess things up here if the TLS section was less aligned than TCB. The offset of a single-byte TLS section would be `-1`, but the start of allocated memory would be at `-alignof(tbc_t)`. However, when `tls_align >= alignof(tbc_t)` then we get
```
size_t alloc_size = ALIGN_UP(tls_size, tls_align) + sizeof(tcb_t);
void *data = alloc(tls_align, alloc_size);
tcb_t *tcb = (tcb_t *) (data - _tcb_data_offset(elf));
```
where `_tcb_data_offset(elf) == -ALIGN_UP(tls_size, tls_align)`

Here, `_tcb_data_offset(elf)` is the same offset compiler computes as well, as per the specification.
However, make it `tls_align < alignof(tbc_t)` and that's no longer the case.

le-jzr: So, yeah. You found an actual bug here! :) It's just such a fringe case we'd have to try really hard to trigger it.

mvolfik: yes, I agree with everything now :+1: Thanks a lot for debugging this mental puzzle with me!

Just a few extra questions:

- this kinda means that the alignment is truly placed **after** the TLS data, just before tcb_t, right?
- isn't this scheme quite weird and brittle, because it depends on the fact that the compiler and the OS compute _tcb_data_offset the same? I.e. that they both agree on the size of the empty space between TLS and the thread pointer?
- and all of this worked so far (and would most likely continue to work forever :D) just thanks to the fact that we always had at least one thread_local structure with alignment >= alignof(tcb_t)?

Also,
> TCB is not part of the ABI

or is it? [Drepper](https://www.akkadia.org/drepper/tls.pdf), section 3.4.1: _The size of the TCB is 16 bytes where the first 8 bytes contain [...], and other 8 bytes are reserved for the implementation_

le-jzr: > yes, I agree with everything now 👍 Thanks a lot for debugging this mental puzzle with me!
> 

You're welcome! Sorry about the false start, thought I could get away with removing my mistake before you started replying. 😅 In retrospect I should have just edited the incorrect comments. Now github's all messed up. 😕  

> Just a few extra checksum questions:
> 
>   * this kinda means that the alignment is truly placed **after** the TLS data, just before tcb_t, right?
> 

For the padding needed for tls_align inside a module, yes. For the multi-module version, or for the above case with small alignment, there might be some padding before data in addition to that.

>   * isn't this scheme quite weird and brittle, because it depends on the fact that the compiler and the OS compute _tcb_data_offset the same? I.e. that they both agree on the size of the empty space between TLS and the thread pointer?

Well, they do need to compute the offset identically. Otherwise the OS wouldn't know where to put the data for compiler-generated code.

> Why not just include this computed padding size in the TLS memsize...

Excellent question. Likely because variant I aligns modules in a different way, so it would waste more memory there, but having that assumption in one variant and not the other would be confusing.

> 
>    * and all of this worked so far (and would most likely continue to work forever :D) just thanks to the fact that we always had at least one thread_local structure with alignment >= alignof(tcb_t)?

Seems like it! :D

> 
> Also,
> 
> > TCB is not part of the ABI
> 
> or is it? [Drepper](https://www.akkadia.org/drepper/tls.pdf), section 3.4.1: _The size of the TCB is 16 bytes where the first 8 bytes contain [...], and other 8 bytes are reserved for the implementation_

That's variant I.
The one with data before thread pointer is variant II.
Why are there two variants you ask? Compatibility with ancient systems, of course, the root of all evil in this world. 😄

mvolfik: > In retrospect I should have just edited the incorrect comments. Now github's all messed up. 😕

welp, lesson learned. Who would guess it breaks like this anyways

I pushed a commit that updates the comments, this thread can be considered closed and PR ready for merge, I hope.

le-jzr:
Nice work, thanks!

PR #241, opened 2025-01-25
ext4: support bigger blkdev block size

mcimerman:
Allows ext4 to be used with block devices that have block size > EXT4_SUPERBLOCK_SIZE (1K).

Currently, ext4 "supports" only block dev bsize smaller or equal to the superblock size. (quotes on support because it gets EPERM when to writing past the allocated buffer)

This currently fails:
```
# mkfile -s 100M /tmp/file1

# /srv/bd/file_bd -b 4096 /tmp/file1 disk1

# mkext4 disk1
```

But the block devices must be at most the block size of the filesystem, currently maximum is 4096 which is the default too.


le-jzr:
Nice. Out of curiosity, do you think there's any value in device blocks even being exposed in the interface like that?
IMO it would be simpler and more efficient to present block devices like normal random access files and let the driver take care of the gritty details.

mcimerman:
> do you think there's any value in device blocks even being exposed in the interface like that?

Do you mean `/srv/bd/file_bd -b 4096 /tmp/file1 disk1`? I think the file backed block device is great for debugging.

le-jzr:
I mean the API interface between a file system and a block device. Like, why should ext4 driver even care what the physical block size is?

mcimerman:
Yes, I get you.

That would make indeed make a lot of things a lot simpler. And it doesn't actually sound that hard to realize.

le-jzr:
> Or is there some pitfall I don't see?

I can't think of anything, but you seem to be more intimately familiar with the code right now, so thought I'd ask you. 😄 

Anyway, right now I'm trying to finish implementing a completely new way to do IPC, and if that works out like I want, then we'll have the opportunity to rework some IPC protocols as they are being migrated to the new shiny thing. Good reason to think about how the seams between layers should work and where there are opportunities for improvement.

le-jzr:
> Edit: Well, probably the block caching will have to be hidden inside the file API, which might not be very easy to do, and it's quite nice that filesystems can manipulate the blocks on a lower level, set some flags or mark them dirty, etc. 

The comparison to files was just illustrative. Obviously it would still be its own protocol with its own nuances, but ability to read/write with arbitrary granularity (and size) like files do would be nice.

Though you make a good point. I'm not actually familiar with the extent of operations that can be done on blocks. But now that I think about it, "trim" would be a good example? Since that can only work on whole blocks, and definitely sounds like something we want to support.

mcimerman:
> seem to be more intimately familiar with the code right now, so thought I'd ask you

I've been working on HelenRAID for a few months now, but in there I use the direct read/write of blocks, just forwarding the final writes or reads that come to the array. Just now I started to look at filesystems, because I would like to do some performance evaluation on "real" workload, with files and whatnot. The filesystems use libblock more extensively, with caching and everything, which I don't fully understand as well as the filesystems.

> new way to do IPC

Sounds great, also the "non-redundant" copying would be great to have, at least for block devices and big IO. I am interested in the new IPC you are working on, do you have any notes or something I can read to try to understand your new proposal? I am not very experienced in that area, but there is a chance I can help :-D

le-jzr:
No formal proposal as of yet, since I'm still figuring out how to implement what I want and details keep changing to accomodate. :)

But currently, the broad strokes look like this: Task creates an IPC queue, then it creates a uniquely identified IPC endpoint on that queue (possibly one for each unique resource handled by the server, e.g. each open file). The endpoint makes its way to a client via IPC (similar to what IPC_CONNECT_*_* calls do now, but with more control and granularity). The client makes calls on the endpoint, providing its own endpoint as a return address.

The main sticking point is that endpoints represent individual resources managed by servers and can be transferred between tasks arbitrarily. They basically act like references to OOP objects that live on a server.

Anyway, the messages sent to endpoint are fixed size, so to send a random bucket of data, you explicitly write the data to a kernel buffer (immutable once created), then send a reference to that buffer to the other party, who can then read chunks out of it. The IPC forward call function is replaced simply by being able to pass the received buffer reference to another task without touching it. Possibly more than once for different parts.

uspace/lib/ext4/src/superblock.c
```diff
@@ -1215,14 +1217,27 @@ errno_t ext4_superblock_write_direct(service_id_t service_id, ext4_superblock_t
...
-	if (EXT4_SUPERBLOCK_SIZE % phys_block_size)
-		block_count++;
+	size_t block_count;
+	if (phys_block_size > EXT4_SUPERBLOCK_SIZE) {
+		block_count = 1;
+		tmp_sb = calloc(1, phys_block_size);
+		if (tmp_sb == NULL)
+			return ENOMEM;
+		void *sb_pos = (uint8_t *)tmp_sb + EXT4_SUPERBLOCK_OFFSET;
+		memcpy(sb_pos, sb, EXT4_SUPERBLOCK_SIZE);
```
le-jzr, line 1227: Don't we need to read the original contents of the physical block first to avoid clearing a few other FS blocks?

le-jzr: On another thought, wouldn't you need to change a bunch of other places in the code to make the filesystem work?

mcimerman: > avoid clearing a few other FS blocks?

Well, the mkfs doesn't care if there was already something or not.

> change a bunch of other places in the code to make the filesystem work?

This is local to just writing the superblock, the reading of the superblock is done via block_read_bytes_direct() which takes care of everything. If you mean the literal working of the filesystem, reading/writing blocks, the cache layer deals with that as well. Or did you have something else on your mind?

le-jzr: > > avoid clearing a few other FS blocks?
> 
> Well, the mkfs doesn't care if there was already something or not.

This is a library. You're assuming nobody will ever ever ever use this function for anything other than mkfs.

le-jzr: Also, this code only works if EXT4_SUPERBLOCK_OFFSET is a multiple of physical block size, which it is not.

mcimerman: > > > avoid clearing a few other FS blocks?

> > Well, the mkfs doesn't care if there was already something or not.

> This is a library. You're assuming nobody will ever ever ever use this function for anything other than mkfs.

Yeah sorry, I meant the function doesn't care.

le-jzr: Why doesn't it care? It's a function that writes the superblock to disk. If the function also wipes stuff that's not the superblock, that seems like a bug and potential issue for any tool that needs to write superblock in an existing filesystem.

le-jzr: > > change a bunch of other places in the code to make the filesystem work?
> 
> If you mean the literal working of the filesystem, reading/writing blocks, the cache layer deals with that as well.

Fair enough.

mcimerman: I see what you mean now. To preserve the bytes before, if there was mbr or gpt or whatever. I am going to fix that, thanks.

le-jzr: Not to mention the function is also used every time the filesystem is mounted or unmounted. With 4096 physical blocks, you'd be wiping logical blocks 0, 2, and 3 every time. With bigger physical blocks, even more would be wiped every time.

mcimerman: Fixed it.

uspace/lib/ext4/src/superblock.c
```diff
@@ -1215,14 +1217,33 @@ errno_t ext4_superblock_write_direct(service_id_t service_id, ext4_superblock_t
...
+	}
 
 	/* Write data */
-	return block_write_direct(service_id, first_block, block_count, sb);
+	rc = block_write_direct(service_id, first_block, block_count, tmp_sb);
+
+	if (phys_block_size > EXT4_SUPERBLOCK_SIZE)
+		free(tmp_sb);
+	return rc;
```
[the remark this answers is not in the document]
le-jzr: Ok, finished. I think this flows nicer. :)

PR #242, opened 2025-01-26
init RTLD runtime at load time even for statically linked binaries

mvolfik:
Another TLS issue, surprisingly completely unrelated to #240

This is about running static binaries, but with CONFIG_RTLD enabled

## What's the current issue

Let's look at the loader code:  
(just keep the code links for reference, I explain the sequence below)

https://github.com/HelenOS/helenos/blob/eff458dc0610e2e236a23436d7688091d3843237/uspace/srv/loader/main.c#L288-L317

The `elf_load` call:
https://github.com/HelenOS/helenos/blob/eff458dc0610e2e236a23436d7688091d3843237/uspace/lib/c/generic/elf/elf_load.c#L57-L87

So, the when the loader is started and runs ldr_load, it first loads the ELF with `elf_load`. After loading the file, for static binaries, `elf_load` simply fills `prog_info.env = NULL`, and `rtld_prog_process` is not called.

So afterwards, back in `ldr_load`, `env == NULL` means that `tls_make` will be called. Let’s look at `tls_make`:

https://github.com/HelenOS/helenos/blob/eff458dc0610e2e236a23436d7688091d3843237/uspace/lib/c/generic/thread/tls.c#L162-L172

So it just defers to `rtld_tls_make` anyways. But this time, with `runtime_env` (a static (=global) variable defined in `lib/c/generic/rtld/rtld.c`. But remember – we are still running the loader, not the program itself. So this runtime_env is the env of the loader, and it contains a pointer to the loader's ELF. So this will create a TLS with the parameters that the loader wants, not the program itself. This is clearly wrong.

---

Later, at startup time, there is this code in `__libc_main`:

https://github.com/HelenOS/helenos/blob/eff458dc0610e2e236a23436d7688091d3843237/uspace/lib/c/generic/libc.c#L97-L104

`__pcb->rtld_runtime` is set to `elf_info->env` in function `elf_set_pcb`, which is called in the `ldr_load` snippet at the top. So the rtld_runtime will be NULL here, and the static initialization is called. This sounds kind of right - but even though this will read the parameters for the TLS and fill it into the environment, `rtld_tls_make` won't be called anymore. This TLS will only be used for any further fibrils. For the main fibril, the TLS+TCB was already created by the broken `tls_make()` call in the loader, and it is used just a few lines earlier in `__libc_main`:

https://github.com/HelenOS/helenos/blob/eff458dc0610e2e236a23436d7688091d3843237/uspace/lib/c/generic/libc.c#L91

(`main_fibril.tcb = __pcb->tcb` earlier, and recall `pcb.tcb = tls_make()` earlier in the loader)

Bonus link:  
how I feel after explaining all of this  
image

## The solution

`rtld_init_static` needs to run in the loader whenever possible (to then correctly create TLS with rtld).

However, it also needs to be able to run in __libc_main, to initialize the runtime even for tasks that don't go through the loader.

## A bit of history

After looking into the history, I found out that `rtld_init_static` was added in its current form in 153c7a29e3a8082bd7b08d69177e32d48d1873a7. At this time, TCB was created in __libc_main even for dynamically linked binaries, so this was probably an obvious choice.

Then later in 40abf56ae4263cdef7e6b8d2a4f5f794c32dedf3, the creation of TCB/TLS was moved from __libc_main to loader, but initialization of static binaries was forgotten here.

Therefore, looking at it with this perspective, this PR just finishes the incomplete movement of the functionality, just over 6 years later.

### DIFF
--- a/uspace/app/dltest/dltest.c
+++ b/uspace/app/dltest/dltest.c
@@ -943,7 +943,7 @@ static int test_dlfcn(void)
 	if (!test_dlfcn_read_public_uvar())
 		return 1;
 
-#ifndef STATIC_EXE
+#ifndef STATIC_EXE // FIXME: this define is not set anywhere
 
 	if (!test_dlfcn_dl_get_private_fib_var())
 		return 1;
--- a/uspace/lib/c/generic/elf/elf_load.c
+++ b/uspace/lib/c/generic/elf/elf_load.c
@@ -64,7 +64,7 @@ errno_t elf_load(int file, elf_info_t *info)
 
 	rc = elf_load_file(file, 0, finfo);
 	if (rc != EOK) {
-		DPRINTF("Failed to load executable '%s'.\n", file_name);
+		DPRINTF("Failed to load executable.\n");
 		return rc;
 	}
 
@@ -72,7 +72,7 @@ errno_t elf_load(int file, elf_info_t *info)
 	DPRINTF("- prog dynamic: %p\n", finfo->dynamic);
 	rc = rtld_prog_process(finfo, &env);
 	if (rc != EOK) {
-		DPRINTF("Failed to process executable '%s'.\n", file_name);
+		DPRINTF("Failed to process executable.\n");
 		return rc;
 	}
 	info->env = env;
--- a/uspace/lib/c/generic/elf/elf_load.c
+++ b/uspace/lib/c/generic/elf/elf_load.c
@@ -56,10 +56,7 @@
  */
 errno_t elf_load(int file, elf_info_t *info)
 {
-#ifdef CONFIG_RTLD
-	rtld_t *env;
-#endif
-	errno_t rc;
+	errno_t rc = EOK;
 
 	rc = elf_load_file(file, 0, &info->finfo);
 	if (rc != EOK) {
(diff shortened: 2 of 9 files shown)


mvolfik:
Updated the PR with some historical context (which makes it clear that I chose the wrong solution), will push updated code in a moment.

mvolfik:
@le-jzr I'm not so sure. IMO `tls_make` is correct, unless you call it inside the special environment of the loader AND rtld is active. If we change it, now every call (from libc or from application code) needs to be aware RTLD, and think about what environment to pass. Also, what argument would we pass in a `! CONFIG_RTLD` environment?

(The only other place where we currently call tls_make is `fibril_alloc`, i.e. for new fibrils except the main one. So here we already have the runtime_env correctly set (or (!), for kernel-loaded programs, runtime_env is NULL and we call the static tls implementation)

mvolfik:
Btw, looking at these two functions again, I think we should merge rtld-initialization of static and dynamic into one function, with some conditional function calls for dynamic. This duplication (placed in different files) seems quite brittle.

image_2025-01-27_15-57-38

le-jzr:
@mvolfik Yeah, you're probably right.

mvolfik:
@le-jzr I think this is ready for review. Tested mainly with `dltest`, which passes on both `ia32` and `amd64` with dynamic linking fully enabled. With RTLD enabled, but `link against shared libraries` disabled, this is the result:

image

I guess this is kind of OK? Because the statically linked libc is initialized, and then the `ldopen`'d library tries to call `tls_get_addr` from a second, dynamically linked `libc`, which breaks. It didn't pass on master before this PR either. It even seems that tests using fibril_local storage should be disabled for static executable builds, but with an `#ifdef` flag that isn't defined anywhere...

With rtld disabled, system boots and simple commands work - this PR shouldn't touch any such code...

Given the above, I assume I didn't break anything, but still, is there any automated CI that would catch any breakage that would be more subtle than this?

PR #243, opened 2025-02-20
fix incorrect syscall usage in tester/mapping1 test

BohdanQQ:
In the `/app/tester`'s `mapping1` test case, `NULL` is passed into the `as_get_physical_mapping` syscall.
This results in the syscall correctly returning `EPERM`, as down the call stack the return value is determined by

```c
// uspace_dst is our incoming NULL
rc = !memcpy_to_uspace(uspace_dst, src, size) ? EPERM : EOK;
```
 (as `memcpy_to_uspace` returns `0` on failure) and the (unexpected) failure of the test case.

As this use of the syscall (passing `NULL` as the only `out` argument) is not reflective of real usage, 
I only adjusted the call site. I also fixed a comment in the assembly which seems incorrect according 
to System V x86-64 ABI.

### DIFF
--- a/kernel/arch/amd64/src/asm.S
+++ b/kernel/arch/amd64/src/asm.S
@@ -72,7 +72,7 @@ FUNCTION_BEGIN(memcpy_to_uspace)
 	rep movsb               /* copy the rest byte by byte */
 
 	0:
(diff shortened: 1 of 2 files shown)


le-jzr:
Thanks!

PR #245, opened 2025-03-06
implement key-based pthread local storage

mvolfik:
This implements `libxposix/pthread/keys.c`. Running destructors requires some fibril exit hooks, so this also adds such API.

As noted in the code comments, the implementation is not fully correct - a) it only supports maximum of 100 keys, and b) there is a rare, but possible race condition when one fibril calls pthread_key_delete, and a fibril in some other thread simultaneously runs the exit hook. I consider this implementation to be a good tradeoff between simplicity and supporting the API for 99% of programs, but if you disagree, I guess I could solve this properly.


le-jzr:
Also, your `pthread_key_delete()` does nothing useful. You can safely make it no-op with no downside and it avoids one of the races. (There's still a race between the exit hook and `pthread_key_create()`. You'll want to fix that, even if it technically works, it's still undefined behavior and it's better not to tempt GCC to f*ck you over with really clever optimizations.)

le-jzr:
> > pthread_key_delete does nothing useful
> 
> That's not true. The [pthread man page](https://linux.die.net/man/3/pthread_key_delete) says that after key delete, destructors for that key shall no longer be called. That is ensured by setting the destructor to NULL

Hmm, right, gonna have to think about it some more.

> 
> > there's still a race
> 
> hm, how? Is short-circuiting evaluation of `&&` not guaranteed or what?

C memory model is far more strict about what it considers a race. Any pair of non-atomic concurrent memory accesses to the same address where at least one of them is a write is *bad*. Since you write non-atomically to the destructor array, and that write can be executed concurrently with a read of the same element in another thread, that's a race and undefined behavior in C land.

> 
> > there shouldn't be exit hooks for fibrils
> 
> I understand that this addition is quite arbitrary.
> 
> Full pthread support is not really my goal (note that the rest of pthread is still completely unimplemented in helenos) - I just needed thread-local storage to get Rust running in HelenOS, and using the pthread-based implementation (which already exists in Rust stdlib) and filling in just this small part of pthread was the easiest option, and also what Vojtech Horky suggested to me. However, none of the standard library uses of `thread_local!` require destructors. Is merging this without support for destructors a reasonable option?

I think so, yeah. After some more thinking I realized there's gotta be some sort of TLS destructor support in libc anyway, since C++ needs it, but I wanna think some more about how to best do it. Also, note that this is my personal opinion and doesn't reflect on any kind of consensus.

le-jzr:
> C memory model is far more strict about what it considers a race.

To put it another way, there's two kinds of races. A "race condition" is what you're thinking of, when different interleaving of the "macroscopic" C operations have different end result.

But there's also a "data race", where memory is being concurrently accessed without synchronization. This is the worse kind, because as far as the standard is concerned, any program where this happens has no semantics. Same as dereference of invalid pointer, or signed integer overflow. It just must not happen, otherwise the program can and will do abhorrent stuff that doesn't make any sense.

le-jzr:
> > > pthread_key_delete does nothing useful
> > 
> > That's not true. The [pthread man page](https://linux.die.net/man/3/pthread_key_delete) says that after key delete, destructors for that key shall no longer be called. That is ensured by setting the destructor to NULL
> 
> Hmm, right, gonna have to think about it some more.

Thinking completed.

So, that requirement for destructors no longer being called basically has two reasons. The first is that it can be used when unloading a DSO (shared library). Unloadable shared libraries are evil, but the idea is that the deinitialization will remove the key with its destructor before unloading the library, because the function is gone at that point and calling it would crash the process.

This sounds like a memory leak machine though, because those destructors aren't called. That's where the other part of intended function comes in. The deinitialization code needs to have a custom cleanup for all the data for that key in all threads. So there's gonna be some kind of list maintained by the library, and the deinit code would first delete the key, then walk the list to deallocate all currently active resources, then unload the DSO.

Now, for either of these to work reliably, it's not enough to just clear the destructor pointer. You need to make sure every destructor call that has been started has completely finished before `pthread_key_delete()` returns. Your code doesn't ensure that, so it's a crash hazard that cannot be safely used for its intended purpose.

The trivial solution would be to just *not* implement that function at all. I doubt Rust std library needs it, since there's no reason for it to want to unload.

le-jzr:
As for the libc side, I've come to the conclusion that it's unavoidable for proper support of thread_local variables in C++, but I don't like adding an API for generic termination hooks if we can avoid it.

Instead, I propose implementing C11 `<threads.h>` in libc, or at least the `tss_create()`, `tss_set()`, `tss_get()` parts for now. To be honest, I'd like to completely replace the fibril API with the C11 standard API. Fibrils are just how threads are implemented in our libc. There's no need to use a unique API just to flaunt our uniqueness. ^_^

mvolfik:
> But there's also a "data race" ...

right, understood, but this just a very elaborate way to tell me "the pointer assignment must be atomic, and you should be more careful about this, this is not just pedantry", am I reading it correctly?

anyways, I removed the destructors and key_delete to at least get this minimal useful patch through

le-jzr:
> > But there's also a "data race" ...
> 
> right, understood, but this just a very elaborate way to tell me "the pointer assignment must be atomic, and you should be more careful about this, this is not just pedantry", am I reading it correctly?

It is *kind of* pedantry, but it is pedantry on the specification's part, that gives leeway to compiler optimizations, and compilers have a habit of unleashing horrible evil whenever the words "undefined" and "behavior" get within a hundred meters of each other, so it's best to respect that pedantry. I might have some residual trauma from too many years of using C. :)) But yes, absolutely. As a general rule, if a mutable object can be touched by two different threads, it has to be either atomic or synchronized by mutexes of some sort. There's exceptions, but those are all black magic.

> anyways, I removed the destructors and key_delete to at least get this minimal useful patch through

Nice. :)

uspace/lib/c/generic/libc.c
```diff
@@ -80,6 +80,7 @@ void __libc_main(void *pcb_ptr)
 
 		main_fibril.tcb = tls_make_initial(__progsymbols.elfstart);
 	}
+	main_fibril.is_freeable = false;
```
le-jzr, line 83: Where did this come from?

uspace/lib/posix/meson.build
```diff
@@ -29,6 +29,7 @@
 
 includes += include_directories('include/posix', 'include')
 c_args += [ '-fno-builtin', '-D_XOPEN_SOURCE' ]
+allow_shared = true
```
le-jzr, line 32: Did you test this with coastline?

uspace/lib/c/generic/rtld/symbol.c
```diff
@@ -64,6 +64,11 @@ static elf_word elf_hash(const unsigned char *name)
 
 static elf_symbol_t *def_find_in_module(const char *name, module_t *m)
 {
+	if (m->dyn.hash == NULL) {
```
le-jzr, line 67: Technically should be a separate PR, but eh, whatever. Curious, did you happen to encounter a crash here?

uspace/lib/c/generic/libc.c
mvolfik: some drive-by changes, as I was digging into fibril startup/exit because of hooks. Does the explicit assignment hurt something? I guess the zero (= false) value is there by default anyways... But it's a counterpart to this code: https://github.com/HelenOS/helenos/blob/0ae9e18465809a5520c51c76be2866b19c48bd0e/uspace/lib/c/generic/thread/fibril.c#L176

uspace/lib/posix/meson.build
mvolfik: 'coastline'?

but I did test it with Rust programs, and also test-libposix passes with dynamic linking enabled

uspace/lib/c/generic/rtld/symbol.c
mvolfik: yep, with Rust programs. This was in the beginnings of the project, I don't really remember. I later had to configure the compiler to create the standard hash section, not the .gnu.hash extension. So maybe the binary that crashed was invalid anyways. Should I try to reconstruct it?

uspace/lib/c/generic/libc.c
le-jzr: Right, as far as I can tell the change does nothing, so I wondered what the rationale is. It doesn't seem like a helpful addition.

uspace/lib/c/generic/rtld/symbol.c
le-jzr: Nah, you don't have to. More robust code never hurts.

uspace/lib/posix/meson.build
le-jzr: "Coastline" is the home of ports that libposix was made for. Separate repository in the HelenOS project. I'll give it a try tomorrow, to make sure there's no weirdness.

le-jzr: To be fair, I think all libraries should be shared anyway. Dunno if there are any lingering issues or if it's just the matter of prudence.

mvolfik: Yeah, I want to look into this deeper in a separate PR. I just wanted libposix for my Rust programs, but I guess I could also just stay with static link until I figure out the uspace/lib/meson.build rewrite

uspace/lib/c/generic/libc.c
mvolfik: I'm really not a fan of implicit zero-initialization as a valid value (other than "missing", i.e. NULLptr), but whatever

le-jzr (approved):

le-jzr:
Works for now. I've already written a proper full implementation for C11 `tss_*` functions, because it was interesting, so I'll swap out the impl once I merge that. :)

PR #249, opened 2025-04-16
lib/c/arch/arm32: add __atomic_{load,store}_8

mcimerman:
GCC automatically expands the use of `ldrd` instruction to use 2 in/out registers, same with `strd`.

Tested on `qemu-system-arm -machine integratorcp`.


le-jzr:
LGTM, thanks!

mcimerman:
Thank you.

PR #251, opened 2025-04-22
uspace/app/blkdump: fix argv null deref

mcimerman:
Previous if conditions might increment `argv`, and the next conditions can dereference the dereference `*(*argv)` (in`str_cmp`) without checking.

Reproduce crash: for example: `blkdump --offset 16`


le-jzr:
Thanks!
At first I thought it's a regression due to me changing `str_cmp()`, but turns out the previous implementation didn't handle null either.

PR #254, opened 2025-05-01
mips32, ppc32: add atomic 8 byte load store, 1 byte cmpexchg

mcimerman:
For now use a mutex.


mcimerman:
Anybody? Should I split it into 2 commits or is this ok?

le-jzr:
> Anybody? Should I split it into 2 commits or is this ok?

I think it's fine as is. No complaints from me, just don't wanna merge it without at least building it first, and this end of semester is nightmarish...

mcimerman:
Sure take your time, this is not urgent, its just for HelenRAID to compile in the future when its in the tree. Good luck with school.

PR #255, opened 2025-05-02
add pointer cursor

mvolfik:
lil chunky boi

image

(unfortunately adding onhover effect to pbutton isn't very simple, because it doesn't store a pointer to the window anywhere (and the set cursor call requires *ui_window_t))


mvolfik:
Well, apparently this is a UI convention specific to web development - majority of buttons on the internet are styled with `cursor:pointer` (StackExchange discussion [1](https://ux.stackexchange.com/questions/3788/default-cursor-on-mouse-over-of-a-button-is-not-a-hand-pointer) [2](https://ux.stackexchange.com/questions/105024/why-dont-button-html-elements-have-a-css-cursor-pointer-by-default)) - and I didn't realize that in desktop environments, this is almost never used.

Is there interest in merging at least the cursor itself (so applications can use it)? Otherwise, I'll close this.

le-jzr:
> apparently this is a UI convention specific to web development

Yeah, makes sense there since every button looks different on the web, so you need an additional way to tell people "you can click on this". Desktop UIs put more focus on visual uniformity.

That said, I'm not sure why the cursor looks like a kid's first mspaint image. Is it supposed to be ironic?

PR #259, opened 2025-05-20
meson/cross/mips32eb: fix compiler name

mcimerman:
When **Big Endian** mips32 is selected in configure.sh, the script tries to find compiler named `mips-helenos-gcc` which is not installed by `toolchain.sh`. Instead it should be `mipsel-helenos-gcc`, the Little Endian and MSIM targets have correct compiler name set.

```
ERROR: $CROSS_PREFIX is not defined and mips-helenos-gcc is not present in any of the following standard locations.
```

### DIFF
--- a/meson/cross/mips32eb
+++ b/meson/cross/mips32eb
@@ -15,7 +15,7 @@ ld = '@COMPPREFIX@ld'
 has_function_printf = true
 needs_exe_wrapper = true
 
-cc_arch = 'mips'
+cc_arch = 'mipsel'
 bits = 32
 
 [host_machine]


le-jzr:
No, that's correct. Big and little endian MIPS toolchains are compiled separately. mipsel is the little-endian target. You might need to run `tools/toolchain.sh mips32eb`

mcimerman:
Yeah sorry I was too fast, got confused with their names didn't think that there is separate toolchain target needed for mips32eb.

le-jzr:
> Yeah sorry I was too fast, got confused with their names and didn't think that there is separate toolchain target needed for mips32eb.

No problem. It is somewhat confusing. :)

mcimerman:
I always thought that `mipsel` is "slang" like term for `mips` :-D

That little endian is taken too literally (`el`)... even for `eb`.

PR #265, opened 2026-08-14
Add CLOCK_MONOTONIC, implemented using uptime

mvolfik:
According to docstring, `getuptime` guarantees monotonicity, so we can use it to implement `CLOCK_MONOTONIC` from `posix/time.h`. Monotonic clock is useful, and needed for Rust support. (I had this code in my branch for a while, but figured I could merge this upstream.)


le-jzr (approved):
lgtm 👍

